PR #25, opened 2016-05-22
Implement the named properties object for global

Sebmaster:


lib/constructs/interface.js
```diff
@@ -26,6 +26,35 @@ Interface.prototype.implements = function (source) {
...
+
+  if (this.idl.inheritance) {
+    this.str += `${this.name}PropertiesConstructor.prototype = Object.create(${this.idl.inheritance}.interface.prototype);
+${this.name}PropertiesConstructor.prototype.constructor = ${this.name}PropertiesConstructor;\n`;
+  }
+
+  this.str += `
+var ${this.name}Properties = new Proxy(${this.name}PropertiesConstructor.prototype, {
+  get(target, key) {
+    if (key in this) {
```
domenic: You should be looking at the target, not `this`

lib/constructs/interface.js
```diff
@@ -26,6 +26,35 @@ Interface.prototype.implements = function (source) {
...
+  throw new TypeError("Illegal constructor");
+}\n`;
+
+  if (this.idl.inheritance) {
+    this.str += `${this.name}PropertiesConstructor.prototype = Object.create(${this.idl.inheritance}.interface.prototype);
+${this.name}PropertiesConstructor.prototype.constructor = ${this.name}PropertiesConstructor;\n`;
+  }
+
+  this.str += `
+var ${this.name}Properties = new Proxy(${this.name}PropertiesConstructor.prototype, {
```
domenic: const

lib/constructs/interface.js
Sebmaster: Yeah, I've only paid attention to the prototype chain for now.

lib/constructs/interface.js
```diff
@@ -26,6 +26,35 @@ Interface.prototype.implements = function (source) {
   this.mixins.push(source);
 };
 
+Interface.prototype.generateNamedPropertiesObject = function () {
+  if (!utils.isGlobal(this.idl)) {
+    return;
+  }
+
+  this.str += `var ${this.name}PropertiesConstructor = function () {
```
domenic: As far as I can tell there's no constructor for the named properties object. It's just a singleton object. So the proxy should have as its target an empty object (probably an `Object.create(null)` object)

lib/constructs/interface.js
```diff
@@ -64,9 +93,12 @@ Interface.prototype.generateConstructor = function () {
 }\n`;
   }
 
-  if (this.idl.inheritance) {
+  if (this.idl.inheritance && !utils.isGlobal(this.idl)) {
     this.str += `${this.name}.prototype = Object.create(${this.idl.inheritance}.interface.prototype);
 ${this.name}.prototype.constructor = ${this.name};\n`;
+  } else if (utils.isGlobal(this.idl)) {
+    this.str += `${this.name}.prototype = Object.create(${this.name}Properties);
```
domenic: Do `Object.setPrototypeOf(obj, namedPropertiesObject)`. Actually the above inheritance should probably do that too (although in that case you'll also need `Object.setPrototypeOf(Class, SuperClass)`.

lib/constructs/interface.js
Sebmaster: You can do `window.__proto__.__proto__.constructor` and it'll return a function which throws `Illegal constructor`.

domenic: Sure, that's EventTargetPrototype.constructor

domenic: Or I guess it's WindowPrototype.constructor

lib/constructs/interface.js
```diff
@@ -26,6 +26,35 @@ Interface.prototype.implements = function (source) {
...
+
+  this.str += `
+var ${this.name}Properties = new Proxy(${this.name}PropertiesConstructor.prototype, {
+  get(target, key) {
+    if (key in this) {
+      return this[key];
+    }
+    return target[key];
+  },
+  set(target, key, value) {
```
domenic: Per spec we shouldn't use the set trap here. Instead it should be the defineProperty and delete traps both of which should return false.

lib/constructs/interface.js
Sebmaster: Oh, that proxy object inherits the EventTarget constructor. Huh!

domenic:
In general we need to run tests to make sure adding a proxy in the global prototype chain doesn't massively slow down everything.

The best benchmark I've found is to clone tc39/ecma262, do `time node_modules/.bin/ecmarkup--verbose spec.html out/index.html`, then force install a version of jsdom patched appropriately, and do it again. We shouldn't regress by more than a few seconds.

PR #26, opened 2016-05-28
Redo API surface

Sebmaster:
Transformations are now wholistic - it's not requried to transform every file by
itself. This means we can access metadata of the whole IDL infrastructure, which
will be required for named properties (due to the named properties visibility
algorithm).


lib/transformer.js
```diff
@@ -0,0 +1,228 @@
...
+      throw new TypeError("idl path has to be a string");
+    }
+    if (typeof impl !== "string") {
+      throw new TypeError("impl path has to be a string");
+    }
+    this.sources.push({ idlPath: idl, impl });
+    return this;
+  }
+
+  *_collectSources() {
```
domenic: IMO the better way to do these is to have them be promise-returning functions, not generator functions. So: `_collectSources: Promise.coroutine(function() { ... })`.

lib/transformer.js
```diff
@@ -0,0 +1,228 @@
...
+    const stats = yield Promise.all(this.sources.map((src) => Promise.fromNode((next) => fs.stat(src.idlPath, next))));
```
domenic: I'd suggest using `require("pn/fs")` instead of promisifying at the call site

package.json
```diff
@@ -2,8 +2,9 @@
   "name": "webidl2js",
   "version": "4.3.3",
   "description": "Auto-generates class structures for WebIDL specifications",
-  "main": "index.js",
+  "main": "lib/transformer.js",
   "dependencies": {
+    "bluebird": "^3.3.5",
```
domenic: co + native promises seems a bit better IMO

lib/transformer.js
```diff
@@ -0,0 +1,228 @@
...
+        relativeUtils = "./" + relativeUtils;
+      }
+
+      source = `"use strict";
+
+const conversions = require("webidl-conversions");
+const utils = require("${relativeUtils}");\n${source}
+const Impl = require("${implFile}.js");\n`;
+
+      fs.writeFileSync(path.join(outputDir, obj.name + ".js"), source);
```
domenic: Don't do sync file writes if you're trying to promise an async interface... same below on line 212

lib/transformer.js
```diff
@@ -0,0 +1,228 @@
...
+              break;
+            }
+            interfaces[instruction.target].implements(instruction.implements);
+            break;
+        }
+      }
+    }
+  }
+
+  *_writeFiles(outputDir) {
```
domenic: Actually this function is just full of sync file reads and writes. Seems bad?

lib/transformer.js
```diff
@@ -0,0 +1,228 @@
...
+      for (const instruction of file.idl) {
+        switch (instruction.type) {
+          case "dictionary":
+            customTypes.add(instruction.name);
+            break;
+        }
+      }
+    }
+
+    for (const file of parsed) {
```
domenic: Maybe a comment explaining this three-stage approach would be nice?

lib/transformer.js
```diff
@@ -0,0 +1,228 @@
...
+    })));
+  }
+
+  *_parse(outputDir, contents) {
+    const parsed = contents.map((content) => ({
+      idl: webidl.parse(content.idlContent),
+      impl: content.impl
+    }));
+
+    const interfaces = this._interfaces = {};
```
domenic: Object.create(null) for map-like objects? Or use a Map proper?

lib/transformer.js
domenic: Or I guess maybe bluebird's promisifyAll is the way to go since you're not using the standard fs module

lib/transformer.js
Sebmaster: Illegal syntax in ES :(

domenic: Aha, right :(

domenic: I guess modify them after the class declaration?

Sebmaster: :( That's not great either. Since it's just a private, I'd probably leave as-is until we get async/await.

lib/transformer.js
Sebmaster: I think we don't actually _need_ a 3 stage approach since nothing should actually kick of a generation during 1st and 2nd stage, so removed "stage 1" and commented the other two.

Sebmaster:
Got everything I think.

lib/transformer.js
```diff
@@ -0,0 +1,225 @@
...
+      }
+    }
+    return files;
+  }
+
+  *_readFiles(files) {
+    const zipped = [];
+    for (const file of files) {
+      zipped.push({
+        idlContent: yield fs.readFile(file.idlPath, { encoding: 'utf-8' }),
```
domenic: This is serial, not parallel, which is vaguely sad.

domenic:
LGTM besides one optional opportunity for more parallelism :D

PR #31, opened 2016-11-10
Implement pair iterators

TimothyGu:
This is incomplete in the sense that it only supports pair iterators through iterable declaration (not value iterators, indexed property getter, maplike, or setlike); and it doesn't specifically handle mixins, overloading, etc. But I think this is a good start, and should cover a large swath of actual iterable API usage.

### DIFF
--- a/lib/constructs/interface.js
+++ b/lib/constructs/interface.js
@@ -5,6 +5,7 @@ const path = require("path");
 const utils = require("../utils");
 const Attribute = require("./attribute");
 const Constant = require("./constant");
+const Iterable = require("./iterable");
 const Operation = require("./operation");
 const Overloads = require("../overloads");
 const Parameters = require("../parameters");
@@ -18,6 +19,7 @@ function Interface(idl, opts) {
   this.requires = {};
   this.str = null;
   this.opts = opts;
+  this.iterable = false;
 }
 
 Interface.prototype.type = "interface";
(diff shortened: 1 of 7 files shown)


lib/constructs/interface.js
```diff
@@ -172,6 +219,13 @@ Interface.prototype.generateExport = function () {
     this.setup(obj, constructorArgs, privateData);
     return utils.implForWrapper(obj);
   },
+  createDefaultIterator(target, kind) {
+    const iterator = Object.create(IteratorPrototype);
```
Sebmaster: Can we just `new` this up normally? I haven't checked the spec for how the object has to be structured.

TimothyGu: The spec just says that the iterator object must have the prototype object as `[[Prototype]]`, but it does not say how the object is to be constructed. It doesn't say anything about `constructor` property, so I guess we could make a wrapper function in the form of

```js
function Iterator(target, kind) {
  this[utils.iterTargetSymbol] = target;
  this[utils.iterKindSymbol] = kind;
  this[utils.iterIndexSymbol] = 0;
}
Iterator.prototype = Object.create(IteratorPrototype);
// note `Iterator.prototype.constructor` is not set
```

That should allow us to get rid of the `createDefaultIterator` also, though if that would be desirable is another question.

The main question is why not `Object.create`?

domenic: I think Object.create is more correct actually, as it's not supposed to have a `.constructor`.

TimothyGu:
@Sebmaster most comments are now fixed.

TimothyGu:
Ping.

domenic:
Oh, thanks for the ping; I forgot to check this repo in my jsdom sweep this last weekend. Hopefully can get to it soon :)

domenic:
Let's merge this. It only handles entries(), not keys() and values() right? Hopefully we can do that in a follow-up.

lib/constructs/iterable.js
```diff
@@ -0,0 +1,67 @@
...
+  if (typeof callback !== "function") {
+    throw new TypeError("Failed to execute 'forEach' on '${this.obj.name}': The callback provided as parameter 1 is not a function.");
+  }
+  const thisArg = arguments[1];
+  let pairs = Array.from(this[impl]);
+  let i = 0;
+  while (i < pairs.length) {
+    const [key, value] = pairs[i];
+    callback.call(thisArg, value, key, this);
+    pairs = Array.from(this[impl]);
```
domenic: I wonder if it would work to just use for-of over the existing one, instead of snapshotting each time?

Also, shouldn't it be `this[impl]()`

lib/constructs/interface.js
```diff
@@ -130,6 +132,52 @@ Interface.prototype.generateExport = function () {
...
 
+  if (this.iterable) {
+    this.str += `\nconst IteratorPrototype = Object.create(utils.IteratorPrototype, {
+  next: {
+    value: function next() {
+      const internal = this[utils.iterInternalSymbol];
+      const target = internal.target;
+      const kind = internal.kind;
+      const index = internal.index;
+      const values = Array.from(target[impl]);
```
domenic: This needs to not snapshot each time. See https://github.com/whatwg/url/issues/188. I will push a fixup commit.

domenic: Oh, nope, it works fine, as long as you snapshot *each time* next() is called. This implementation is inefficient, but we can tidy that up later.

TimothyGu: This is really just a bare-bones implementation of the spec. Not very efficient for anything that is not stored as an array in the impl, but the spec does ask for a "list of value pairs to iterate over", which in ECMAScript should be expressed as an array.

One thing that could be optimized is the `switch` statement involving `kind`. We could create three functions for each `kind` value and then store the callback pointer in the internal object

PR #32, opened 2016-12-22
Add [WebIDL2JSFactory] for producing an interface factory function	

TimothyGu:


Sebmaster:
Hrm, I kinda dislike putting jsdom specific logic into this package, but I'm not sure if we can avoid it forever 😞

TimothyGu:
Neither do I, but unfortunately there isn't really a better way to solve this problem (see incomplete alternative in tmpvar/jsdom#720b1021cf0b6a4aee90cee68005669c160918b2)

domenic:
I guess we could have some sort of plugin system and have a separate package for jsdom-specific parts of webidl2js. But that's definitely overkill for now.

I think we might over time want to make this more general: about factories, not just about _ownerDocument and EventTarget. But for now this gets the job done.

lib/constructs/interface.js
```diff
@@ -55,6 +56,15 @@ Interface.prototype.generateConstructor = function () {
     }
     this.str += conversions.body + "\n";
 
+    if (utils.getExtAttr(this.idl.extAttrs, "JSDOMEventTarget")) {
+      this.str += `
+  Object.defineProperty(this, "_ownerDocument", {
```
Sebmaster: This will expose `_ownerDocument` on the wrapper, yes? We definitely want to avoid that. Can we just provide it to the impl in the constructor?

TimothyGu: I guess that would work.

domenic:
LGTM, but I kind of feel we should name nonstandard extended attributes specially? E.g. [JSDOMFactory] ?

Sebmaster:
I wonder if it makes sense to make this a general kind of thing without the extended Attr? Like have a method on exports which gives you a new object every time which automatically provides a set of args to `privateData`, and also provide a default object if you don't need that ability?

TimothyGu:
@domenic, changed to `[W2JSFactory]` to make this less JSDOM-specific.

@Sebmaster, that sounds like a much bigger project than this would require. However, I do think the ability to specify _any_ `privateData`, not just `window`, to be beneficial, so I changed the argument to `createInterface` to `defaultPrivateData`.

domenic:
I pushed some minor nits. This looks good, however, what I didn't realize until I looked at some generated output is that this pushes *everything* onto the factory. So you do

```js
const eventTargetFactory = require("./generated/EventTarget");
const interface = eventTargetFactory();

dom.EventTarget = interface.interface;

// and now you have to use interface.is, interface.create, etc.
```

This is a bit tricky thinking about how we'd use this in jsdom. I guess we'd store the `interface` object on each Window's impl? Hmm, but wouldn't this break generated brand checks, if we ever get those---like

```js
const isEventTarget = require("./generated/EventTarget").is;

// ...later...
if (!isEventTarget(args[0])) {
  throw new TypeError("...");
}
```

This seems bad.

Would there be a way to instead generate things so that is/isImpl are still exported and still work correctly across all instances? After all that better mimics Web IDL's semantics, where brand-checking works even across realms.

TimothyGu:
> Would there be a way to instead generate things so that is/isImpl are still exported and still work correctly across all instances?
At first I tried lifting the prototype object out of factory function, which would make `instanceof` work automatically, but then I realized each created prototype object needs to have a different `constructor` property.

> After all that better mimics Web IDL's semantics, where brand-checking works even across realms.

Hmm. I don't really see a Node.js v0.12-compliant way around this except adding a per-interface symbol to every created object…

Is it really needed for an EventTarget implementation to have access to something akin to `_ownerDocument`?

TimothyGu:
@domenic, ping.

domenic:
> Hmm. I don't really see a Node.js v0.12-compliant way around this except adding a per-interface symbol to every created object…

We only need Node.js v4 compliance, if that helps. Adding a per-interface symbol seems pretty reasonable to me...

> Is it really needed for an EventTarget implementation to have access to something akin to _ownerDocument?

The design isn't great. It's used in a few places in EventTarget-impl:

- [Checking if the event originates from a "closed Window"](https://github.com/tmpvar/jsdom/blob/cbf50d4ac69fbd534ce3b4d3caff6033d9c374d9/lib/jsdom/living/events/EventTarget-impl.js#L176-L180). Events firing on closed windows has caused jsdom problems in the past, so we went with this solution, although I think there are probably better ways of doing this in a spec-compliant manner.
- [Trying to find where to report an exception](https://github.com/tmpvar/jsdom/blob/cbf50d4ac69fbd534ce3b4d3caff6033d9c374d9/lib/jsdom/living/events/EventTarget-impl.js#L222-L236)
- [Checking if inline event listeners are happening in a "closed Window"](https://github.com/tmpvar/jsdom/blob/cbf50d4ac69fbd534ce3b4d3caff6033d9c374d9/lib/jsdom/living/events/EventTarget-impl.js#L162)

All of these could be likely replaced with some kind of associated Window, not Document. But we still need something like that, that tells us the EventTarget instance's associated Window, for these cases.

TimothyGu:
@domenic, turns out I didn't need a symbol, as `is` and `isImpl` check the implementation (if any) of the incoming object. PTAL.

Sebmaster:
I'll check it out tonight.

lib/constructs/interface.js
```diff
@@ -325,7 +348,25 @@ Interface.prototype.generate = function () {
 
   this.generateSymbols();
 
-  this.generateExport();
+    this.str += `
```
TimothyGu: Fixed. Thanks for noticing.

TimothyGu:
> This change is also incompatible with iterables. Not sure if it's needed, but definitely a caveat to note somewhere.

Fixed.

Issue #29, opened 2016-07-02
Handling overloads


domenic:
Would be nice if given

```
void append(USVString name, USVString value);
void append(USVString name, Blob value, optional USVString filename);
```

I could write `append1()` and `append2()` in the impl class and have it call to the correct one.

TimothyGu:
Using a numeric index can be very volatile, and also doesn't play well when you want to call the impl method from another impl class. Also, in general the [overload resolution algorithm](https://heycam.github.io/webidl/#dfn-overload-resolution-algorithm) specifies runtime semantics, and is pretty difficult to translate to static semantics. E.g. for

```
void append(USVString name, Node arg1, Blob arg2);
void append(USVString name, Blob arg1, Node arg2);
void append(USVString name, Error arg1);
```

webidl2js must ideally generate something like

```js
if (Blob.is(args[1])) {
  args[1] = unwrap(args[1]);
  if (Node.is(args[2])) {
    args[2] = unwrap(args[2]);
  } else {
    throw new TypeError("Argument is not a Node");
  }
} else if (Node.is(args[1])) {
  args[1] = unwrap(args[1]);
  if (Blob.is(args[2])) {
    args[2] = unwrap(args[2]);
  } else {
    throw new TypeError("Argument is not a Blob");
  }
} else if (args[1] instanceof Error) {
  args[1] = conversions["Error"](args[1]); // or omitted
} else {
  throw new TypeError("Argument is not a Blob or a Node");
}
```

which IMO is a pretty big feat to do statically.

Alternatively, we could merge the IDL above into

```
void append(USVString name, (Node or Blob or Error) arg1, (Blob or Node) arg2);
```

which is a lot easier to generate:

```js
// Group all interface types together when generating (step 4)
// https://heycam.github.io/webidl/#es-to-union

// arg1
if (Node.is(args[1]) || Blob.is(args[1])) { // step 4
  args[1] = unwrap(args[1]);
} else if (args[1] instanceof Error) { // step 6
  args[1] = conversions["Error"](args[1]); // or omitted
} else { // step 17
  throw new Error("Argument is not Node, Blob or Error");
}

// arg2
if (Node.is(args[2]) || Blob.is(args[2])) { // step 4
  args[2] = unwrap(args[2]);
} else {
  throw new Error("Argument is not Node or Blob");
}
```

The flip-flop error checking logic can be implemented easily enough in the impl class (and this error checking is good to have for the cases when other impl's call this method themselves anyway):

```js
[...]

Issue #5, opened 2015-09-27
Unroll argument conversion loop (for non-... args)


domenic:
``` js
  const args = [];
  for (let i = 0; i < arguments.length && i < 3; ++i) {
    args[i] = arguments[i];
  }
  args[0] = conversions["DOMString"](args[0]);
```

could become

``` js
args[0] = conversions["DOMString"](arguments[0]);
args[1] = arguments[1];
args[2] = arguments[2];
```

This also helps with #4 since it automatically fills out `args[2]` with `undefined` if it wasn't passed.

Sebmaster:
I guess this is kinda related to #4. I'm still hanging on to that thought of ensuring arguments.length equivalence for impl classes.

domenic:
I am pretty sure impl classes should not rely on arguments.length though. I don't think WebIDL even allows that checking. It is all handled in the WebIDL layer.

Sebmaster:
In order to fix the dictionary thing, `arguments.length` has been broken already, so we won't need a major for further changes there, but I haven't unrolled the loop yet since we need to do a bit of conditionals there to support variadic args.

domenic:
I notice this kind of happened but there's a tryImplForWrapper loop above the unrolled conversions now. I think we should be able to apply tryImplForWrapper only to (known?) interface types.

Sebmaster:
> I think we should be able to apply tryImplForWrapper only to (known?) interface types.

Not right, now, since we never know about all the interface types. Take Node.dispatchEvent for example. We don't actually know about the Event type there (in jsdom, due to our folder structure at least, since we convert all folders seperately).

I plan to update the conversion interface to include more state so you can add multiple folders to the same conversion process, so we'd know about the whole interface system, but haven't gotten to it yet.

TimothyGu:
An update: since about six months ago we no longer try to get the impl if the type is known not to be an interface (sequences, records, promises, primitives). There is still the major pain point that we don't yet have a way to import types from other modules. E.g. the URL interface is implemented in whatwg-url using webidl2js, but the webidl2js in jsdom doesn't know URL is an interface, and even though it still tries unwrapping it doesn't work because the impl symbol is different for interfaces in another module.

I've got a fix in the works, but will have to change up the type system a bit here so that it doesn't assume the impl symbol is the same for all interfaces. I'll also have to think of a solution for hybrids like Window, in that it functions like a generated class and has impl property (because it inherits EventTarget), but is not actually generated.

PR #33, opened 2017-02-06
Type conversion overhaul

TimothyGu:
Implementing union type conversion is next, followed by overload resolution algorithm (if need be).

### DIFF
--- a/lib/constructs/dictionary.js
+++ b/lib/constructs/dictionary.js
@@ -33,7 +33,7 @@ class Dictionary {
     if (value !== undefined) {`;
       const argAttrs = field.extAttrs;
 
-      const conv = Types.generateTypeConversion(this.ctx, "value", typeConversion, argAttrs, this.name);
+      const conv = Types.generateTypeConversion(this.ctx, "value", typeConversion, argAttrs, this.name, `\`\${context} has member ${field.name} that\``);
       for (let key in conv.requires) {
         this.str = `const ${key} = ${conv.requires[key]};\n` + this.str;
       }
@@ -63,26 +63,23 @@ class Dictionary {
   generate() {
     this.str += `
 module.exports = {
-  convertInherit(obj, ret) {`;
+  convertInherit(obj, ret, { context = "The provided value" } = {}) {`;
     if (this.idl.inheritance) {
       this.str = `const ${this.idl.inheritance} = require("./${this.idl.inheritance}");\n` + this.str;
       this.str += `
-    ${this.idl.inheritance}.convertInherit(obj, ret);`;
+    ${this.idl.inheritance}.convertInherit(obj, ret, { context });`;
     }
     this._generateConversions();
     this.str += `
   },
 
-  convert(obj) {
+  convert(obj, { context = "The provided value" } = {}) {
     if (obj !== undefined && typeof obj !== "object") {
-      throw new TypeError("Dictionary has to be an object");
-    }
-    if (obj instanceof Date || obj instanceof RegExp) {
-      throw new TypeError("Dictionary may not be a Date or RegExp object");
+      throw new TypeError(\`\${context} is not an object.\`);
     }
 
     const ret = Object.create(null);
-    module.exports.convertInherit(obj, ret);
+    module.exports.convertInherit(obj, ret, { context });
     return ret;
   }
 };`;
--- a/lib/constructs/dictionary.js
+++ b/lib/constructs/dictionary.js
@@ -33,7 +33,7 @@ class Dictionary {
     if (value !== undefined) {`;
       const argAttrs = field.extAttrs;
 
-      const conv = Types.generateTypeConversion(this.ctx, "value", typeConversion, argAttrs);
+      const conv = Types.generateTypeConversion(this.ctx, "value", typeConversion, argAttrs, this.name);
       for (let key in conv.requires) {
         this.str = `const ${key} = ${conv.requires[key]};\n` + this.str;
       }
--- a/lib/constructs/dictionary.js
+++ b/lib/constructs/dictionary.js
@@ -4,10 +4,10 @@ const Types = require("../types");
 const utils = require("../utils");
 
 class Dictionary {
-  constructor(idl, opts) {
+  constructor(ctx, idl) {
+    this.ctx = ctx;
     this.idl = idl;
     this.name = idl.name;
-    this.opts = opts;
   }
 
   _generateConversions() {
@@ -33,7 +33,7 @@ class Dictionary {
     if (value !== undefined) {`;
       const argAttrs = field.extAttrs;
 
-      const conv = Types.generateTypeConversion("value", typeConversion, argAttrs, this.opts.customTypes);
+      const conv = Types.generateTypeConversion(this.ctx, "value", typeConversion, argAttrs);
       for (let key in conv.requires) {
         this.str = `const ${key} = ${conv.requires[key]};\n` + this.str;
       }
--- a/lib/constructs/dictionary.js
+++ b/lib/constructs/dictionary.js
@@ -74,7 +74,7 @@ module.exports = {
   },
 
   convert(obj, { context = "The provided value" } = {}) {
-    if (obj !== undefined && typeof obj !== "object") {
+    if (obj !== undefined && typeof obj !== "object" && typeof obj !== "function") {
       throw new TypeError(\`\${context} is not an object.\`);
     }
 
--- a/lib/constructs/dictionary.js
+++ b/lib/constructs/dictionary.js
@@ -1,6 +1,6 @@
 "use strict";
 
-const conversions = require("webidl-conversions");
+const Types = require("../types");
 const utils = require("../utils");
 
 class Dictionary {
@@ -32,34 +32,15 @@ class Dictionary {
       this.str += `
     if (value !== undefined) {`;
       const argAttrs = field.extAttrs;
-      const enforceRange = utils.getExtAttr(argAttrs, "EnforceRange");
-      const clamp = utils.getExtAttr(argAttrs, "Clamp");
 
-      let optString = "";
-      if (clamp) {
-        optString = `, { clamp: true }`;
-      } else if (enforceRange) {
-        optString = `, { enforceRange: true }`;
+      const conv = Types.generateTypeConversion("value", typeConversion, argAttrs, this.opts.customTypes);
+      for (let key in conv.requires) {
+        this.str = `const ${key} = ${conv.requires[key]};\n` + this.str;
       }
+      this.str += conv.body;
+      this.str += `
+    ret[key] = value;`
 
-      let conversionFn = "";
-      if (conversions[typeConversion.idlType]) {
-        conversionFn = `conversions["${typeConversion.idlType}"]`;
-      } else if (this.opts.customTypes.has(typeConversion.idlType)) {
-        this.str = `const convert${typeConversion.idlType} = require("./${typeConversion.idlType}");\n` + this.str;
-        conversionFn = `convert${typeConversion.idlType}`;
-      }
-
-      if (typeConversion.array) {
-        this.str += `
-      ret[key] = [];
-      for (let i = 0; i < value.length; ++i) {
-        ret[key][i] = ${conversionFn}(value[i]${optString});
-      }`;
-      } else {
-        this.str += `
-      ret[key] = ${conversionFn}(value${optString});`;
-      }
       if (field.required) {
         this.str += `
     } else {
--- a/lib/transformer.js
+++ b/lib/transformer.js
@@ -7,20 +7,20 @@ const fs = require("pn/fs");
 const webidl = require("webidl2");
 const prettier = require("prettier");
 
+const Context = require("./context");
 const Interface = require("./constructs/interface");
 const Dictionary = require("./constructs/dictionary");
 
 class Transformer {
-  constructor(opts) {
+  constructor(opts = {}) {
+    this.ctx = new Context({
+      implSuffix: opts.implSuffix
+    });
     this.options = Object.assign({
-      implSuffix: "",
       suppressErrors: false
     }, opts);
 
     this.sources = [];
-    this._interfaces = null;
-    this._dictionaries = null;
-    this._customTypes = null;
   }
 
   addSource(idl, impl) {
@@ -76,9 +76,8 @@ class Transformer {
       impl: content.impl
     }));
 
-    const interfaces = this._interfaces = Object.create(null);
-    const dictionaries = this._dictionaries = Object.create(null);
-    const customTypes = this._customTypes = new Set();
+    this.ctx.initialize();
+    const { interfaces, dictionaries, customTypes } = this.ctx;
 
     // first we're gathering all full interfaces and ignore partial ones
     for (const file of parsed) {
@@ -90,10 +89,8 @@ class Transformer {
               break;
             }
 
-            obj = new Interface(instruction, {
-              implDir: path.resolve(outputDir, file.impl),
-              implSuffix: this.options.implSuffix,
-              customTypes
+            obj = new Interface(this.ctx, instruction, {
+              implDir: path.resolve(outputDir, file.impl)
             });
             interfaces[obj.name] = obj;
             break;
@@ -104,7 +101,7 @@ class Transformer {
               break;
             }
 
-            obj = new Dictionary(instruction, { customTypes });
+            obj = new Dictionary(this.ctx, instruction);
             dictionaries[obj.name] = obj;
             customTypes.add(instruction.name);
             break;
@@ -162,23 +159,25 @@ class Transformer {
     let utilsText = yield fs.readFile(__dirname + "/output/utils.js");
     yield fs.writeFile(this.options.utilPath, utilsText);
 
+    const { interfaces, dictionaries } = this.ctx;
+
     let interfaceStub = yield fs.readFile(__dirname + "/output/interfaceStub.js");
-    let keys = Object.keys(this._interfaces).concat(Object.keys(this._dictionaries));
+    let keys = Object.keys(interfaces).concat(Object.keys(dictionaries));
     for (let key of keys) {
-      if (this._interfaces[key]) {
-        yield fs.writeFile(path.join(outputDir, this._interfaces[key].name + ".js"), interfaceStub);
+      if (interfaces[key]) {
+        yield fs.writeFile(path.join(outputDir, interfaces[key].name + ".js"), interfaceStub);
       } else {
-        yield fs.writeFile(path.join(outputDir, this._dictionaries[key].name + ".js"), "");
+        yield fs.writeFile(path.join(outputDir, dictionaries[key].name + ".js"), "");
       }
     }
 
-    keys = Object.keys(this._interfaces);
+    keys = Object.keys(interfaces);
     for (let i = 0; i < keys.length; ++i) {
-      const obj = this._interfaces[keys[i]];
+      const obj = interfaces[keys[i]];
       let source = obj.toString();
 
       const implDir = path.relative(outputDir, obj.opts.implDir).replace(/\\/g, "/"); // fix windows file paths
-      let implFile = implDir + "/" + obj.name + this.options.implSuffix;
+      let implFile = implDir + "/" + obj.name + this.ctx.implSuffix;
       if (implFile[0] !== ".") {
         implFile = "./" + implFile;
       }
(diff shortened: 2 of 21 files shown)


lib/constructs/dictionary.js
```diff
@@ -1,6 +1,6 @@
 "use strict";
 
-const conversions = require("webidl-conversions");
+const Types = require("../types.js");
```
domenic: Since this is not a class, it should not be capitalized. Also, in this project it seems like we're not including `.js` in the requires.

lib/types.js
```diff
@@ -0,0 +1,88 @@
+"use strict";
+
+const conversions = require("webidl-conversions");
+
+const utils = require("./utils");
+
+module.exports.generateTypeConversion = function (name, idlType, argAttrs, customTypes, opts) {
```
domenic: This function is very long. I think it would benefit from calling out to a few non-exported factored-out functions for the two cases of sequence and IDL type. What do you think?

lib/types.js
```diff
@@ -0,0 +1,88 @@
...
+  }
+
+  return {
+    requires,
+    body: str
+  };
+};
+
+module.exports.canHandleType = function (idlType, customTypes) {
+  return idlType.generic === "sequence" || conversions[idlType.idlType] || customTypes.has(idlType.idlType);
```
domenic: `idlType.idlType in conversions` seems a bit nicer since it guarantees the return value will always be a boolean.

lib/types.js
```diff
@@ -38,6 +38,43 @@ module.exports.generateTypeConversion = function (name, idlType, argAttrs, custo
...
+  if (${name} === null && ${name} === undefined) {
+    ${name} = new Map;
+  } else {`;
+    }
+
+    str += `
+  if (typeof ${name} !== "object") {
+    throw new TypeError("The value provided is not an object");
+  } else {
+    const result = new Map;
```
domenic: Nit: parens after constructor

lib/types.js
```diff
@@ -38,6 +38,43 @@ module.exports.generateTypeConversion = function (name, idlType, argAttrs, custo
...
+    const keys = Object.keys(${name});
```
domenic: Per spec this should be Object.getOwnPropertyKeys, and thne you should use getOwnPropertyDescriptor, and check enumerability, and so on. This is important as it impacts how many side effects occur before an exception happens in exceptional cases.

lib/transformer.js
```diff
@@ -162,12 +169,14 @@ class Transformer {
...
     let interfaceStub = yield fs.readFile(__dirname + "/output/interfaceStub.js");
-    let keys = Object.keys(this._interfaces).concat(Object.keys(this._dictionaries));
+    let keys = Object.keys(this._interfaces).concat(Object.keys(this._dictionaries)).concat(Object.keys(this._typedefs));
     for (let key of keys) {
       if (this._interfaces[key]) {
         yield fs.writeFile(path.join(outputDir, this._interfaces[key].name + ".js"), interfaceStub);
-      } else {
+      } else if (this._dictionaries[key]) {
         yield fs.writeFile(path.join(outputDir, this._dictionaries[key].name + ".js"), "");
+      } else {
```
domenic: I guess while here it would be good to change this to `else if (this._typedefs[key])` as that seems a bit less likely to blow up in the future.

lib/constructs/dictionary.js
TimothyGu: I'm mirroring how `require('../parameters')` is usually called `Parameters`, even though `Parameters` is not a class. See e.g. [constructs/interface.js](https://github.com/jsdom/webidl2js/blob/7a70dc7c7f90c16cbb92c880f219fea024436846/lib/constructs/interface.js#L11). Will fix the `.js`.

lib/types.js
TimothyGu: Agreed, will split.

lib/transformer.js
TimothyGu: TBH this entire thing should be changed to use `Map` anyway...

TimothyGu:
> Only major issue is the dictionary conversion is more subtle than done here.

Any specific cases you have in mind?

> This is overall really awesome; thanks for working on it. Would you be able to show some sample outputs? Eventually I guess we should get better about writing tests for webidl2js, but for now just being able to spot-check the output would be helpful.

Yeah of course. I tested this against JSDOM to make sure it still works.

> Are there particular features you're planning on adding to jsdom using these?

Haven't decided yet, though I want to eventually implement the Fetch API and spec. Might also implement `URLSearchParams` (which uses records and union types) for whatwg-url as a proof of concept.

lib/constructs/dictionary.js
domenic: Oh, I see. I guess good to stay consistent, even though that's a weird style choice from @Sebmaster :)

domenic:
> Any specific cases you have in mind?

Oops, I meant records, not dictionaries, as per the comments you already saw.

> Haven't decided yet, though I want to eventually implement the Fetch API and spec. Might also implement URLSearchParams (which uses records and union types) for whatwg-url as a proof of concept.

That would be awesome! +1 to URLSearchParams as a first step; as seen in https://github.com/jsdom/whatwg-url/issues/37 people seem to definitely want this.

TimothyGu:
@domenic, okay, updated. I made some example I/Os in https://github.com/TimothyGu/webidl2js/tree/34c9ae6c1356d7a6ec0a898a774a2479414acc26/test. `cases`  contain the IDL source, while `output` has all the generated JS files.

domenic:
OK, awesome! Are you planning to commit those in this PR?

Reviewing the output:

- [It looks like record args go through the unwrapping](https://github.com/TimothyGu/webidl2js/blob/34c9ae6c1356d7a6ec0a898a774a2479414acc26/test/output/ClassUsingRecord.js#L26-L28). That's not ideal; we should be operating on the wrapper, not the impl. Same for sequence. This may require fixing https://github.com/jsdom/webidl2js/issues/5 :-/.
- Indentation seems off in general, but I think that's better solved by a postprocessing step. Filed https://github.com/jsdom/webidl2js/issues/35.
- The [type checks inside the sequence conversion](https://github.com/TimothyGu/webidl2js/blob/34c9ae6c1356d7a6ec0a898a774a2479414acc26/test/output/ClassUsingSequence.js#L24-L26) are unnecessary, I believe. `for-of` will give the same errors if you pass it a non-object/non-iterable, and technically it's better to not access the `[Symbol.iterator]` property twice.

TimothyGu:
I just pushed a patch that fixes arg unwrapping (and hence #5) but it's fairly big I'll need more time to make sure things like variadic arguments still work, so **DO NOT MERGE**. I've tested it against jsdom though and other than a few small fixes here and there in jsdom, its tests all pass.

Another thing is that I changed my mind regarding typedef. They should not generate interface files, but rather be inlined and flattened when they are used elsewhere. This is necessary because union type handling requires fully flattened member types. I have not fixed this yet, however.

Do you want me to include the test cases? And if so, do you want to just put them in test/ or write some actual tests with them?

Updated outputs are in https://github.com/TimothyGu/webidl2js/tree/354eb7acaa5c2911c895cd858da1f5c8df68fc64/test (compare with [before](https://github.com/TimothyGu/webidl2js/tree/34c9ae6c1356d7a6ec0a898a774a2479414acc26/test).

<details><summary>Here's what's changed (so far):</summary>

- If the argument is of an interface type, check if the argument is indeed of that type and throw an error otherwise
  - Use `this.requires` more extensively to prevent duplicated `require()`'s
  - Pass `interfaces` to individual classes alongside `customTypes`
  - Use an `opts` object containing both in various places
- Instead of unwrapping as soon as an argument is seen, first store the arguments in an array then process them one-by-one
  - Then, unwrap only if needed for interface types or for currently unhandled types (notably, union type)
- Add support for `Promise` and `FrozenArray` types
  - Not necessarily needed in this PR, but to prevent trying to unwrap them when it is not needed
- Handle nullability solely in `Types.generateTypeConversion()` (instead of in both `Parameters` and `Types`), and handle optionality before nullability per spec
  - Remove `Types.canHandleType()`
- Fix default arguments processing
  - None of the default arguments in jsdom use Infinity or NaN, which would have immediately exposed the bug
  - Default argument as null was previously masked by incorrect order of optionality/nullability handling
- Use `Types.generateTypeConversion()` instead of `Parameters.generateVarConversion()` in `Attributes`
  - Attributes setters have nothing to do with parameter optionality, which is now the only thing `Parameters.generateVarConversion()` handles
  - Unexport `Parameters.generateVarConversion()`
- Fix faulty logic in `Overloads.proveSimiliarity()`

</details>

stevenvachon:
+1000 for `URLSearchParams`

domenic:
> Do you want me to include the test cases? And if so, do you want to just put them in test/ or write some actual tests with them?

I'm not sure what our best strategy is here personally. What do you think? In my opinion it's useful to know when things change, so maybe trying to use [baseline-tester](https://www.npmjs.com/package/baseline-tester) would do the trick?

We could at some point try to write semantic tests, e.g. actually pass in various inputs to a generated file, and check how they get converted to a sequence by the time the test impl sees them. That sounds like a lot of work though :-/.

Those fixes are all really wonderful <3. Looking forward to being able to merge this.

> Updated outputs are in https://github.com/TimothyGu/webidl2js/tree/354eb7acaa5c2911c895cd858da1f5c8df68fc64/test 

This still seems to do a loop to copy `arguments` into `args`, right? I had envisioned the loop being fully unrolled, so that you process each `arguments` member in turn, with the effect of the block processing `arguments[i]` being that `args[i]` is set.

domenic:
@TimothyGu thoughts on moving this forward? I'd love to merge it :)

stevenvachon:
@TimothyGu ❓ 👍

TimothyGu:
I'm having some problems with overloaded rest arguments. I haven't had a lot of time to work on this recently, but I'll try to get back at it over this weekend.

stevenvachon:
@TimothyGu hi 😄

TimothyGu:
Guess what… the HD that contained my WIP crashed a few weeks ago, so the commits I didn't push to GitHub are now gone. I'll see what I can do about it, but I'll have to change the implementation for typedef anyway so I might just consider breaking up this PR into smaller chunks…

stevenvachon:
💥 🖥 💥

TimothyGu:
Rebased. I removed the experimental (i.e. buggy) fix for #5 as well as incorrect typedef and union support. Those are bigger projects that will need to be tackled at a later time.

After those changes, this PR now only contains:

- Support for record, sequence, Promise, FrozenArray types
- Support for interface types as first-class types
   - E.g. now calling `Node.prototype.appendChild` with a string will throw an error from the wrapper
- Only unwrap parameters when necessary
   - Arguments are still copied to a new array, however
- Better Chrome-like error context
   - ```
     > window.eval('document.createTreeWalker(document.body).currentNode = 2;')
     TypeError: Failed to set the 'currentNode' property on 'TreeWalker': The provided value is not of type 'Node'.
         at convert (jsdom/lib/jsdom/living/generated/Node.js:490:11)
         at TreeWalker.set [as currentNode] (jsdom/lib/jsdom/living/generated/TreeWalker.js:90:9)
         at eval (eval at <anonymous> (repl:1:8), <anonymous>:1:54)
         at eval (<anonymous>)
         at repl:1:8
     ```
- Simplify attribute generation due to conflation of parameters and types
- Remove outdated special dictionary handling for Date and RegExp
- Subtle fixes for default parameters and overload similarity algorithm
- Internal refactoring

domenic:
Awesome! I guess I'm most able to review this by trying to figure out its impact on jsdom; apart from that, each commit looks reasonable. You say that all the tests are green with that diff? That's awesome and a little surprising :D.

> ```diff
> -  const event = ErrorEvent.createImpl(["error", {
> +  const event = ErrorEvent.create(["error", {
> ```

What are these changes about? In general our philosphy was that impl code should only know about or deal with impls, if possible.

> ```diff
> -//  void append(USVString name, USVString value); // handle overloads manually; https://github.com/jsdom/webidl2js/issues/29
> +  void append(USVString name, USVString value);
>    void append(USVString name, Blob value, optional USVString filename);
> ```

What does uncommenting these help with? It still seems like we're handling overloads manually.

> ```diff
> -    if (!HTMLElement.isImpl(value)) {
> -      throw new TypeError("Argument must be a HTMLElement");
> -    }
> -    if (value._namespaceURI !== "http://www.w3.org/1999/xhtml" ||
> +    if (!HTMLElement.isImpl(value) ||
> +        value._namespaceURI !== "http://www.w3.org/1999/xhtml" ||
> ```

Why are we still testing HTMLElement.isImpl?

TimothyGu:
> You say that all the tests are green with that diff?

Yep :smile:

>> ```diff
>> -  const event = ErrorEvent.createImpl(["error", {
>> +  const event = ErrorEvent.create(["error", {
>> ```
>
> What are these changes about? In general our philosphy was that impl code should only know about or deal with impls, if possible.

This is indeed pretty tricky, so I'm going to try to explain it the best I can. The function this piece of diff is located in, [reportAnError()](https://github.com/tmpvar/jsdom/blob/d078830eafd17ec38aa816cf283da39ec5316e31/lib/jsdom/living/helpers/runtime-script-errors.js#L7-L34), is always [called](https://github.com/tmpvar/jsdom/blob/d078830eafd17ec38aa816cf283da39ec5316e31/lib/jsdom/living/helpers/runtime-script-errors.js#L54) with `target` being an instance of Window. The [Window interface](https://github.com/tmpvar/jsdom/blob/d078830eafd17ec38aa816cf283da39ec5316e31/lib/jsdom/browser/Window.js) in JSDOM is implemented by hand, but [extends](https://github.com/tmpvar/jsdom/blob/d078830eafd17ec38aa816cf283da39ec5316e31/lib/jsdom/browser/Window.js#L506) a generated class `EventTarget`. Now, after this PR the generated code enforces strict check for parameter types, which means that only ErrorEvent wrappers are accepted in the [`EventTarget.prototype.dispatchEvent()`](https://github.com/tmpvar/jsdom/blob/d078830eafd17ec38aa816cf283da39ec5316e31/lib/jsdom/living/helpers/runtime-script-errors.js#L29) wrapper, which is what the `event` above is passed to, _not_ the underlying impl function.

Now this does mean that instead of passing a wrapper function around in the guts, we can specifically unwrap `window`, like so:

```diff
diff --git a/lib/jsdom/living/helpers/runtime-script-errors.js b/lib/jsdom/living/helpers/runtime-script-errors.js
index 5f334e87..71f91477 100644
--- a/lib/jsdom/living/helpers/runtime-script-errors.js
+++ b/lib/jsdom/living/helpers/runtime-script-errors.js
@@ -1,5 +1,6 @@
 "use strict";
 const util = require("util");
+const idlUtils = require("../generated/utils");
 const ErrorEvent = require("../generated/ErrorEvent");
 
 const errorReportingMode = Symbol("error reporting mode");
@@ -15,7 +16,7 @@ function reportAnError(line, col, target, errorObject, message, location) {
   target[errorReportingMode] = true;
 
   // TODO Events: use constructor directly, once they are no longer tied to a window.
-  const event = ErrorEvent.create(["error", {
+  const event = ErrorEvent.createImpl(["error", {
     bubbles: false,
     cancelable: true,
     message,
@@ -51,7 +52,7 @@ module.exports = function reportException(window, error, filenameHint) {
   const lineNumber = pieces && parseInt(pieces[3]) || 0;
   const columnNumber = pieces && parseInt(pieces[4]) || 0;
 
-  const handled = reportAnError(lineNumber, columnNumber, window, error, error.message, fileName);
+  const handled = reportAnError(lineNumber, columnNumber, idlUtils.tryImplForWrapper(window), error, error.message, fileName);
 
   if (!handled) {
     const errorString = shouldBeDisplayedAsError(error) ? `[${error.name}: ${error.message}]` : util.inspect(error);
```

I can do that instead if that's what you prefer.

----

I'll explain the other changes later today.

TimothyGu:
>> ```diff
>> -//  void append(USVString name, USVString value); // handle overloads manually; https://github.com/jsdom/webidl2js/issues/29
>> +  void append(USVString name, USVString value);
>>    void append(USVString name, Blob value, optional USVString filename);
>> ```
>
> What does uncommenting these help with? It still seems like we're handling overloads manually.

Yep we are (for now), but if the first line is not commented webidl2js will think Blob is the only possible type for the second parameter, and therefore throws an error if it is not.

>> ```diff
>> -    if (!HTMLElement.isImpl(value)) {
>> -      throw new TypeError("Argument must be a HTMLElement");
>> -    }
>> -    if (value._namespaceURI !== "http://www.w3.org/1999/xhtml" ||
>> +    if (!HTMLElement.isImpl(value) ||
>> +        value._namespaceURI !== "http://www.w3.org/1999/xhtml" ||
>> ```
>
> Why are we still testing HTMLElement.isImpl?

The old code is actually not spec-compliant, since by [spec](https://html.spec.whatwg.org/multipage/dom.html#dom-document-body) *value* can legally be either `null` or a `HTMLElement`, and in case of `null` a "`HierarchyRequestError`" `DOMException` should be thrown instead of `TypeError`. Now, non-null and non-HTMLElement *value*s are filtered in the IDL layer, so there is no need for the continued existence of `throw new TypeError` here. But we still need to handle the case where *value* is `null`.

domenic:
> Now this does mean that instead of passing a wrapper function around in the guts, we can specifically unwrap window, like so:

Thanks for the explanation. Yeah, I think passing the "Window impl" (actually an EventTarget impl, if I understand correctly) would be the way to go here. That also has the benefit that we don't get borked when some on-page script overwrites `EventTarget.prototype.dispatchEvent`.

It might be worth adding a comment to `reportAnError` that it takes an EventTarget impl, to help things be clearer. (It's time like these one can appreciate the benefits of TypeScript...)

> Yep we are (for now), but if the first line is not commented webidl2js will think Blob is the only possible type for the second parameter, and therefore throw an error if it is not a Blob.

Hmm. Looking at the generated output though, I don't see it working great...

```js
URLSearchParams.prototype.append = function append(name, value) {
  if (!this || !module.exports.is(this)) {
    throw new TypeError("Illegal invocation");
  }
  if (arguments.length < 2) {
    throw new TypeError("Failed to execute 'append' on 'URLSearchParams': 2 arguments required, but only " + arguments.length + " present.");
  }

  const args = [];
  for (let i = 0; i < arguments.length && i < 2; ++i) {
    args[i] = arguments[i];
  }
  args[0] = conversions["USVString"](args[0], { context: "Failed to execute 'append' on 'URLSearchParams': parameter 1" });
  args[1] = conversions["USVString"](args[1], { context: "Failed to execute 'append' on 'URLSearchParams': parameter 2" });
  return this[impl].append(...args);
};
```

Should we instead be creating our own IDL, like

```
void append(USVString name, any value, optional USVString filename);
```

? In any case this warrants commenting in the IDL file.

I guess this also explains the impl -> wrapper change

> ```diff
>    append(name, value, filename) {
>      // Handling this manually for now: https://github.com/jsdom/webidl2js/issues/29
> -    if (!Blob.isImpl(value)) {
> +    if (Blob.is(value)) {
> +      value = Blob.convert(value);
> +    } else {
>        value = conversions.USVString(value);
>      }
> ```

You removed all the tryWrapperToImpl stuff, so now we need to assume unknown inputs are wrappers. But again it seems like we should be using `any`...

> Now, non-null and non-HTMLElement values are filtered in the IDL layer, so there is no need for the continued existence of throw new TypeError here. But we still need to handle the case where value is null.

Can we just test for null then?

---

Overall it sounds like the webidl2js side of all this is pretty solid, so probably not many changes needed in this PR, but I thought it'd be good to be sure before merging.

TimothyGu:
> It might be worth adding a comment to reportAnError that it takes an `EventTarget` impl, to help things be clearer. 

Agreed, done.

----

> Hmm. Looking at the generated output though, I don't see it working great...

What you were initially referring to was actually `FormData`, not `URLSearchParams`. Compare:

```diff
--- orig.js	2017-05-06 14:33:21.436963019 -0700
+++ after-pr.js	2017-05-06 14:33:16.084943368 -0700
@@ -8,11 +8,12 @@
 
   const args = [];
   for (let i = 0; i < arguments.length && i < 3; ++i) {
-    args[i] = utils.tryImplForWrapper(arguments[i]);
-  }
-  args[0] = conversions["USVString"](args[0]);
-  if (args[2] !== undefined) {
-  args[2] = conversions["USVString"](args[2]);
+    args[i] = arguments[i];
   }
+  args[0] = conversions["USVString"](args[0], { context: "Failed to execute 'append' on 'FormData': parameter 1" });
+  args[1] = convertBlob(args[1], { context: "Failed to execute 'append' on 'FormData': parameter 2" });
+if (args[2] !== undefined) {
+  args[2] = conversions["USVString"](args[2], { context: "Failed to execute 'append' on 'FormData': parameter 3" });
+}
   return this[impl].append(...args);
 };
```

The important thing here is that `convertBlob()` gets the impl of the provided value is a Blob, and throws a TypeError otherwise. We don't want it to throw here, and that's why I deliberately uncommented the other overload to make sure webidl2js doesn't try to call `convertBlob()`.

> Should we instead be creating our own IDL, like
>
> ```js
> void append(USVString name, any value, optional USVString filename);
> ```
>
>?

Or yeah, this works too. Changed.

> In any case this warrants commenting in the IDL file.

Again, agreed.

> I guess this also explains the impl -> wrapper change

Yep, though I added back `tryImplForWrapper` and support for impl in the impl method.

----

> Can we just test for null then?

Sure.

----

I've created a PR at jsdom with the jsdom changes to ease merging: https://github.com/tmpvar/jsdom/pull/1841

domenic:
All right, let's give this a shot...

PR #36, opened 2017-04-24
Add support for Symbol.toStringTag (closes #21).

### LINKED ISSUE #21: Properly implement toStringTag
[W3C Bug Tracker](https://www.w3.org/Bugs/Public/show_bug.cgi?id=28244) has a discussion going on how to implement it. It's also not supported in a stable node yet, need for v6 for it anyways.

droooney:


TimothyGu:
This isn't actually compliant to the Web IDL spec, which says the prototype's \@\@toStringTag must be the interface's name + `'Proto'`. But from @domenic in https://github.com/nodejs/node/pull/10906#issuecomment-273957809:

> Regarding toStringTag, this is actually a controversial part of the Web IDL spec. See https://www.w3.org/Bugs/Public/show_bug.cgi?id=28244. I would suggest following Chrome's version, of just putting it as a data property on the prototype, but I am somewhat biased there :).

Regardless, you probably should remove the existing `prototype.toString` function.

domenic:
Yes, I am much happier with this than the version in the Web IDL spec, which nobody implements. But yes, we need to remove the toString function that already exists.

droooney:
Should I set a getter according to the specs (add `'Prototype'` for the prototype) then (which seems controversial to me as well)?

domenic:
No, as I said, I am much happier with this version than the version in the Web IDL specs. The only change is that we need to remove the toString function that we're currently installing on every class.

droooney:
No problem :) I personally like this version more as well.

Sebmaster:
We need to keep toString for the stringifier and impl toString method. Only the stringtag part should be removed.

droooney:
Oops, sorry... Ridiculously long commit history for such a small change :)

lib/constructs/interface.js
```diff
@@ -365,18 +365,15 @@ Interface.prototype.generateToString = function () {
     }
   }
   if (!hasToString) {
```
domenic: This condition is now flipped; it should be `if (hasToString)`.

I'll fix this before merging.

domenic:
I ended up doing this a bit differently; it turns out that whole generateToString() function is no longer needed. See https://github.com/jsdom/webidl2js/commit/0590e5a8b3e91edf39d40209795ff41b87787bfe. Thanks for the kick-start though!

droooney:
No problem :) Thanks for awesome jsdom!

PR #37, opened 2017-04-29
Support the [SameObject] extended attribute

Zirro:
Thanks for helping out with the code.


lib/constructs/attribute.js
```diff
@@ -44,6 +45,10 @@ Attribute.prototype.generate = function () {
     setterBody = reflector[this.idl.idlType.idlType].set(objName, attrName);
   }
 
+  if (sameObject) {
+    getterBody = `return utils.getSameObject(this, "${this.idl.name}", () => { ${getterBody} })`;
```
domenic: Add a semicolon before the backtick :)

Zirro: Oops. Fixed :)

PR #38, opened 2017-05-03
Try to automatically wrap iterated key/value

TimothyGu:
See https://github.com/tmpvar/jsdom/pull/1734#discussion_r100679464

A patch for jsdom is available at https://github.com/TimothyGu/jsdom/commit/ce679c3489263c99a4e14a6cd1e93751b141a4df

### DIFF
--- a/lib/constructs/interface.js
+++ b/lib/constructs/interface.js
@@ -20,7 +20,7 @@ function Interface(idl, opts) {
   this.requires = {};
   this.str = null;
   this.opts = opts;
-  this.iterable = this.idl.members.some(member => member.type === "iterable");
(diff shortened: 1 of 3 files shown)


domenic:
LGTM. Would an accurate description of this patch (for the commit message) also include

> Also ensures we don't try to apply pair iterator logic to value iterators, leaving those as more explicitly unimplemented for now.

?

TimothyGu:
Yes, that was more of a drive-by fix than anything. Do you want me to amend the commit or can you handle it yourself?

domenic:
I've got it, just didn't want to misrepresent while merging :)

PR #39, opened 2017-05-03
Make all symbol props non-writable non-enumerable

TimothyGu:
Node.js v8.x starts to display all enumerable symbols in inspection, which means that w/o this change the implementation of all DOM objects will be visible to the user.

A similar change is done to `@@toStringTag` and `@@unscopables` to maintain compatibility with Chrome.

### DIFF
--- a/lib/constructs/interface.js
+++ b/lib/constructs/interface.js
@@ -191,11 +191,12 @@ Interface.prototype.generateExport = function () {
     this.str += `
   createDefaultIterator(target, kind) {
     const iterator = Object.create(IteratorPrototype);
-    iterator[utils.iterInternalSymbol] = {
-      target,
-      kind,
-      index: 0
-    };
+    Object.defineProperty(iterator, utils.iterInternalSymbol, {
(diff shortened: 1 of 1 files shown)


TimothyGu:
This will necessitate some changes on the jsdom side: https://github.com/TimothyGu/jsdom/commit/04a54f6532c7f11dff83f883ff8c7aaa2b2dfba7

In other words this is probably a breaking change.

domenic:
I don't think this is a breaking change, although jsdom should change to be consistent. But jsdom is not going to break because of this, just be a little inconsistent.

TimothyGu:
I ran the jsdom test suite and there will be some failing tests, since several handwritten classes attempt to extend the IDL interfaces, but they do so by setting @@toStringTag directly. And since the property is readonly on the prototype, and they are using strict mode, the setting operation fails

domenic:
Ah, I see; thanks for explaining! I agree then it's a breaking change; I guess I should try to remember in the future that flipping writable to false has such implications.

PR #40, opened 2017-05-06
Add Jest-based snapshot testing and Travis

domenic:
Previously we were just kind of running test/test.js as needed and hoping for the best. With this in place we have an easier way of checking for regressions.

Let's get Travis running on the PR first...

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,13 @@
+language: node_js
+node_js:
+  - 6
+  - stable
+
+branches:
+  only:
+    - master
+
+notifications:
+  email:
+    on_success: never
+    on_failure: always
--- a/package.json
+++ b/package.json
@@ -10,9 +10,12 @@
     "webidl-conversions": "^4.0.0",
     "webidl2": "^2.0.11"
   },
-  "devDependencies": {},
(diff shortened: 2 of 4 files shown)


.travis.yml
```diff
@@ -0,0 +1,16 @@
...
+  only:
+    - master
+
+notifications:
+  email:
+    on_success: never
+    on_failure: always
+
+script:
+  - npm test
```
Sebmaster: That should be the default so not needed.

.travis.yml
```diff
@@ -0,0 +1,16 @@
+language: node_js
+node_js:
+  - 6
+  - stable
+
+branches:
+  only:
+    - master
```
TimothyGu: This shouldn't be needed either. It's not the default but I don't think it's necessary.

Sebmaster: It should reduce the double travis-build in PRs on direct branches.

TimothyGu: Ah never mind then.

.travis.yml
domenic: Yeah I guess I was copying it from somewhere where we also had `npm run lint`

PR #46, opened 2017-07-31
Miscellaneous improvements

TimothyGu:
- Variadic argument handling in "get effective override" imported straight from the spec
- Use Map instead of Object.create(null) for Context#dictionaries/interfaces, something I've wanted to do for quite some time
- Move Impl.init check to interface runtime, to simplify build process and eliminates circular dependencies. This does add a conditional to the generated files, but realistically the performance cost should be minimal.

### DIFF
--- a/lib/constructs/interface.js
+++ b/lib/constructs/interface.js
@@ -297,7 +297,6 @@ Interface.prototype.generateIface = function () {
 
     this._internalSetup(obj);\n`;
 
-  const implClass = require(this.opts.implDir + "/" + this.name + this.ctx.implSuffix);
   this.str += `
     Object.defineProperty(obj, impl, {
       value: new Impl.implementation(constructorArgs, privateData),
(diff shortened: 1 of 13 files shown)


domenic:
I tend to prefer using Object.create(null) for string-keyed maps, as the syntax is more convenient. Why switch?

Very excited about the circular dependency elimination.

TimothyGu:
I was under the impression that `Map`s are designed from the start to handle this sort of tasks, and indeed they perform better than `Object.create(null)` on current and future V8 versions. It's also somewhat easier to iterate on the values of a `Map` w/o universal `Object.values()` support in supported Node.js versions. These reasons are all inconsequential though, so I'd be happy to defer to your judgement.

TimothyGu:
@domenic Would it be possible for me to have write permission to this repo? I am in the jsdom org but not in any teams, nor was I added to any repo as a collaborator. Being able to merge PRs would make development much quicker and easier.

domenic:
@TimothyGu done!

Issue #35, opened 2017-02-08
Reformat the output somehow


domenic:
The indentation is already not perfect, and https://github.com/jsdom/webidl2js/pull/33 is going to add more imperfect indentation.

It would be ideal if while writing webidl2js, we could just ignore the indentation issues (or use whichever looks nicer inside the webidl2js source code), then have a postprocessing step that reformats everything for us. Maybe eslint's fix option, or maybe something simpler based on just esprima.

TimothyGu:
A more radical approach is we can just rewrite the entire code generator based on Babel or escodegen -- that is, generating an AST instead of JS code directly. Something like [babel-template](https://www.npmjs.com/package/babel-template) makes it less painful than it would appear at first.

Sebmaster:
Oh yes. babel-template looks awesome. I originally wanted to avoid building ASTs due to the verbosity, but this might just be perfect.

domenic:
It's worth noting that the original [domenic/webidl-class-generator](https://github.com/domenic/webidl-class-generator) did something similar using Traceur's babel-template counterpart. See e.g. https://github.com/domenic/webidl-class-generator/blob/master/lib/code-generation.js

domenic:
For direct source-to-source prettification, https://github.com/jlongster/prettier is the new hotness.

domenic:
OK, so I did the prettier part of this. That was easy.

Remaining tasks until I personally am happy:

- [ ] Reformat all multiline code strings in the source to be indented in a way that is pleasing to read and write, instead of concerned with what the output would look like. (E.g., have leading whitespace indentation as if they were code and so on.)
- [ ] Add an eslint config to this project, probably similar to jsdom's, and fix all issues with this project.
- [ ] Change from prettier to [prettier-eslint](https://github.com/prettier/prettier-eslint) to get even more cleanups and fixups, using the project's ESLint config.

But I want to delay these until landing #33 as I don't want to make @TimothyGu rebase unnecessarily on any lint fixes and reformatting. So this is on hold for a little bit.

TimothyGu:
> Change from prettier to prettier-eslint to get even more cleanups and fixups

I'm not sure if we want to put ESLint in webidl2js's `dependencies`, which is necessary for prettier-eslint to work. ESLint is much bigger than prettier, while prettier has 0 dependencies (since it's bundled by rollup).

Sebmaster:
[...]

PR #47, opened 2017-08-02
Style and lint

TimothyGu:
Fixes most of #35, doesn't use prettier-eslint yet as I don't think that's necessary.


lib/constructs/interface.js
```diff
@@ -396,7 +397,7 @@ Interface.prototype.generateSymbols = function () {
   if (Object.keys(unscopables).length) {
     this.str += `
       Object.defineProperty(${this.name}.prototype, Symbol.unscopables, {
-        value: ${JSON.stringify(unscopables, null, '  ')},
+        value: ${JSON.stringify(unscopables, null, "  ")},
```
domenic: Huh, I never knew you could do that; I always did `2`.

TimothyGu:
I pulled out the test-adding commit. Now https://github.com/jsdom/webidl2js/pull/47/commits/d5677bdcb9491402715c100853ab81b6a573ec13#diff-e80cf17e46910d548417d47d60230718 should be pristine :ok_hand:

TimothyGu:
Note that the stringifier tests didn't work before this PR (the emitted JS isn't valid JS), and I fixed stringifier code generation in the reformat commit, only after it got too hairy to pull it out :( That was why I originally had the tests with the reformat commit. Should be all good now though.

lib/constructs/interface.js
TimothyGu: Some people prefer tabs :)

PR #48, opened 2017-08-07
Implement legacy platform object + others

TimothyGu:
Companion jsdom PR: https://github.com/tmpvar/jsdom/pull/1934

### DIFF
--- a/lib/constructs/interface.js
+++ b/lib/constructs/interface.js
@@ -16,10 +16,10 @@ class Interface {
     this.name = idl.name;
     this.factory = Boolean(utils.getExtAttr(this.idl.extAttrs, "WebIDL2JSFactory"));
 
-    this.mixins = [];
-    this.requires = {};
     this.str = null;
     this.opts = opts;
+    this.requires = new utils.RequiresMap(ctx);
+    this.mixins = [];
   }
 
   get hasPairIterator() {
@@ -91,7 +91,7 @@ class Interface {
 
       const conversions =
         Parameters.generateOverloadConversions(this.ctx, overloads, this.name, `Failed to construct '${this.name}': `);
-      Object.assign(this.requires, conversions.requires);
+      this.requires.merge(conversions.requires);
 
       minConstructor.nameList = minConstructor.nameList.map(name => (keywords.has(name) ? "_" : "") + name);
       this.str += `
@@ -143,21 +143,21 @@ class Interface {
   }
 
   generateRequires() {
-    this.requires.impl = "utils.implSymbol";
+    this.requires.addRaw("impl", "utils.implSymbol");
 
     if (this.idl.inheritance !== null) {
-      this.requires[this.idl.inheritance] = `require("./${this.idl.inheritance}.js")`;
+      this.requires.add(this.idl.inheritance);
     }
 
     if (this.mixins.length !== 0) {
-      this.requires.mixin = "utils.mixin";
+      this.requires.addRaw("mixin", "utils.mixin");
       for (const mixin of this.mixins) {
-        this.requires[mixin] = `require("./${mixin}.js")`;
+        this.requires.add(mixin);
       }
     }
 
     this.str = `
-      ${Object.keys(this.requires).map(key => `const ${key} = ${this.requires[key]};`).join("\n")}
+      ${this.requires.generate()}
 
       ${this.str}
     `;
@@ -363,7 +363,7 @@ class Interface {
 
       if (member !== null) {
         const data = member.generate();
-        Object.assign(this.requires, data.requires);
+        this.requires.merge(data.requires);
         this.str += data.body;
       }
     }
@@ -391,7 +391,7 @@ class Interface {
 
       if (member !== null) {
         const data = member.generate();
-        Object.assign(this.requires, data.requires);
+        this.requires.merge(data.requires);
         this.str += data.body;
       }
     }
--- a/lib/constructs/interface.js
+++ b/lib/constructs/interface.js
@@ -301,12 +301,12 @@ class Interface {
     this.str += `
       create(constructorArgs, privateData) {
         let obj = Object.create(${this.name}.prototype);
-        this.setup(obj, constructorArgs, privateData);
+        obj = this.setup(obj, constructorArgs, privateData);
         return obj;
       },
       createImpl(constructorArgs, privateData) {
         let obj = Object.create(${this.name}.prototype);
-        this.setup(obj, constructorArgs, privateData);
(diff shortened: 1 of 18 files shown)


lib/constructs/interface.js
```diff
@@ -9,6 +9,36 @@ const Overloads = require("../overloads");
 const Parameters = require("../parameters");
 const keywords = require("../keywords");
 
+function isPropGetter(idl) {
+  return idl.type === "operation" && idl.getter && Array.isArray(idl.arguments) && idl.arguments.length === 1;
+}
+
+// eslint-disable-next-line no-unused-vars
+function isPropSetter(idl) {
```
domenic: Hmm I'd rather not include these if they are dead code...

lib/constructs/interface.js
```diff
@@ -22,8 +52,20 @@ class Interface {
     this.mixins = [];
   }
 
+  get supportsIndexedProperties() {
```
domenic: Since these are relatively expensive operations (O(n) in number of members) I think they should not be getters. Just `iface.supportsIndexedProperties()` makes sense to me.

lib/constructs/iterable.js
```diff
@@ -34,40 +34,20 @@ class Iterable {
...
-          }
-        };
       `;
     } else {
-      // value iterator; WIP
+      str += `
+        ${this.obj.name}.prototype.entries = Array.prototype.entries;
+        ${this.obj.name}.prototype.keys = Array.prototype.keys;
+        ${this.obj.name}.prototype.values = Array.prototype.values;
+      `;
```
domenic: I'm unclear on the split between the two files; why do some operations go here and others in interface.js?

lib/constructs/interface.js
TimothyGu: It won't be dead in the next commit :)

lib/constructs/iterable.js
TimothyGu: `entries`, `keys`, and `values` will only be present on an interface if the interface has an iterable declaration ([source]), while `Symbol.iterator` and `forEach` may be present in other situations [as well](https://heycam.github.io/webidl/#es-iterators). I thought it was more appropriate to put the things directly related to iterable declarations in `iterable.js`.

TimothyGu:
Thank you for the review!

> The four commit should probably mention why you're returning the obj from setup. I assume it is so that you can return a proxy in the next commit?

Yep, I expect the commits to be eventually squashed into one so I didn't explain too much in detail.

lib/constructs/interface.js
TimothyGu: After the next commit they'll be O(1) :)

lib/constructs/interface.js
```diff
@@ -10,33 +10,13 @@ const Overloads = require("../overloads");
...
-
-function isPropDeleter(idl) {
-  return idl.type === "operation" && idl.deleter && Array.isArray(idl.arguments) && idl.arguments.length === 1;
-}
-
-function isNamed(idl) {
-  return idl.arguments[0].idlType.idlType === "DOMString";
-}
+// Used as a sentinel in inheritedMembers to signify that the following members are inherited.
+const kInherited = Symbol("inherited");
```
domenic: We haven't used a "k" prefix on constants before.

lib/constructs/interface.js
```diff
@@ -10,33 +10,13 @@ const Overloads = require("../overloads");
...
-
-function isValueIterable(idl) {
-  return idl.type === "iterable" && !Array.isArray(idl.idlType);
-}
-
-function isPairIterable(idl) {
-  return idl.type === "iterable" && Array.isArray(idl.idlType) && idl.idlType.length === 2;
-}
+/* eslint-disable func-style */
+const isNamed = idl => idl.arguments[0].idlType.idlType === "DOMString";
```
domenic: Can you use the normal function style here as indicated by the lint rules, instead of disabling them?

lib/constructs/interface.js
```diff
@@ -317,11 +446,11 @@ class Interface {
         return `${value} !== ${unsupportedValue}`;
       }
       return `${O}[impl][utils.supportsPropertyName](${P})`;
-    }
+    };
 
     // "named property visibility algorithm"
     // If `supports` is true then skip the supportsPropertyName check.
-    function namedPropertyVisible(P, O, supports = false) {
+    const namedPropertyVisible = (P, O, supports = false) => { // eslint-disable-line func-style
```
domenic: Again please stick with the normal function style; if you're disabling a line rule that's a sign something has gone wrong style-wise.

lib/constructs/interface.js
```diff
@@ -227,6 +416,528 @@ class Interface {
...
+          }
+      `;
+    }
+    this.str += `
+          return Reflect.deleteProperty(target, P);
+        },
+    `;
+
+    // We cannot implement [[Call]] or legacycallers yet, since ES proxies don't let us intercept [[Call]]s to a
+    // non-callable target.
```
domenic: Well, we control the target... but we can leave that for later, certainly.

lib/constructs/interface.js
TimothyGu: Sure, I can change them. I just thought the one-line style is more appropriate for this type of shorthand function.

lib/constructs/interface.js
TimothyGu: Other such functions all use `this` value somewhere, so the linter doesn't complain about it. This one doesn't yet, but it seems odd that the styles are different for functions of the same ilk.

I'll change it of course, just wasn't entirely convinced enabling this style rule for these functions was the best idea.

domenic: Nah, I prefer function declarations for functions, and arrow functions for cases where the function is being passed somewhere.

domenic:
Oh, can you explain [WebIDL2JSValueAsUnsupported]? Why/when is it necessary?

TimothyGu:
Sure. In certain IDL algorithms named properties are used like "if *P* is a supported property name then perform the step of determining the value corresponding to named property *P*". For some jsdom interfaces the process of getting the value can be somewhat time consuming, and in most cases determining if a property is supported is the exact same process as getting the property itself, only returning a boolean instead of the property value. Rather than going through that process (at least) twice on every get, I decided to provide the ability to specify a value to mean "unsupported" (usually null in practice) so that the two roundtrips can be reduced to one. Having that extended attribute would also simplify the implementation class (one less webidl2js-specific method).

domenic:
So, we'll definitely need good documentation for this, after #51 and #60 land. But for now, let me get it merged :D.

PR #65, opened 2017-08-21
Add simple variadic operation support

TimothyGu:
Needed for DOMTokenList.

### DIFF
--- a/test/__snapshots__/test.js.snap
+++ b/test/__snapshots__/test.js.snap
@@ -4512,6 +4512,191 @@ const Impl = require(\\"../implementations/UnforgeableMap.js\\");
 "
 `;
 
+exports[`Variadic.idl 1`] = `
+"\\"use strict\\";
+
+const conversions = require(\\"webidl-conversions\\");
+const utils = require(\\"./utils.js\\");
+
+const convertURL = require(\\"./URL.js\\").convert;
+const impl = utils.implSymbol;
+
+function Variadic() {
+  throw new TypeError(\\"Illegal constructor\\");
+}
+
+Object.defineProperty(Variadic, \\"prototype\\", {
+  value: Variadic.prototype,
+  writable: false,
+  enumerable: false,
+  configurable: false
+});
+
+Variadic.prototype.simple1 = function simple1() {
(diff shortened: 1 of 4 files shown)


test/__snapshots__/test.js.snap
```diff
@@ -4393,6 +4393,191 @@ const Impl = require(\\"../implementations/UnforgeableMap.js\\");
...
+        \\" present.\\"
+    );
+  }
+
+  const args = [];
+  for (let i = 0; i < arguments.length; ++i) {
+    args[i] = arguments[i];
+  }
+
+  if (args[1] !== undefined) {
```
domenic: Why not convert past the argument at index 1 like the other examples do?

test/cases/Variadic.idl
```diff
@@ -0,0 +1,10 @@
+interface Variadic {
+  void simple1(DOMString... strings);
+  void simple2(DOMString first, URL... urls);
```
domenic: Can you add a space between these so that only overloads are grouped together in this file?

TimothyGu:
> The main problem here is the commit message and comment use the term "simple variadic operation" which I am not familiar with.

I meant it as "an un-overloaded operation that has a variadic argument and zero or more regular arguments", in contrast with "overloaded variadic operation", both of which terms I used in the comment in the parameter conversion code. Overloaded operations, with one or more of them supporting variadic arguments, are not supported at this moment.

In other words, the "simple1" and "simple2" cases are the only ones I intend to support right now, while "overloaded1" and "overloaded2" are included so as to make sure they are excluded from the variadic argument handling.

I'll try to make this more explicit.

test/__snapshots__/test.js.snap
TimothyGu: This operation is overloaded. As stated in https://github.com/TimothyGu/webidl2js/blob/6b8788120218090738d815104eaa5db584799f7a/lib/parameters.js#L80-L84 I don't want to support overloaded operations with variadic arguments just yet, especially since jsdom doesn't require it at this moment.

TimothyGu:
@domenic Docs added. Hopefully the language is a bit clearer.

Issue #55, opened 2017-08-19
Don't export most things, for mixins


domenic:
See also #53. is/isImpl and maybe mixedInto make sense. But most of the rest of the exports don't.

To reliably identify a mixin, we can either wait for https://github.com/heycam/webidl/issues/363 or we can use a custom extended attribute in the meantime.

TimothyGu:
Are there examples of interfaces defined with NoInterfaceObject but is not a mixin?

domenic:
A few legacy ones. E.g. https://html.spec.whatwg.org/#external

TimothyGu:
In that case, could we specifically flag the few NoInterfaceObject non-mixin interfaces and default to the mixin behavior for NoInterfaceObject? I could see the argument against this of course but just thought it might cause less churn.

domenic:
I think I'd rather tag the mixins since we're going to eventually make mixins first class in Web IDL anyway so it'd be nice to get a sense of which ones they are.

TimothyGu:
Sounds just fine to me.

ExE-Boss:
Is this supposed to be closed?

Because README.md says that this is still an issue: https://github.com/jsdom/webidl2js/blob/9f103a29bd58a79f8ac7dcf7fe5cac43cd176ed0/README.md#L130

Also, #68’s description says that it doesn’t fix this.

TimothyGu:
Ugh, GitHub's auto-closed this issue even though I said:

> Note, this does not yet fix #55

Thanks for the note @ExE-Boss!

ExE-Boss:
Yeah, that’s because you wrote `“fix”` followed by the issue number separated by a single space (`U+0020`), [which is all GitHub cares about](https://help.github.com/en/articles/closing-issues-using-keywords).

TimothyGu:
Fixed in #135. Interface mixins don’t cause files to be generated.

Issue #56, opened 2017-08-19
convertInherited for dictionaries should be _convertInherited


domenic:
It's not "public" for consumers of the generated wrapper class; it's used by webidl2js only.

Issue #52, opened 2017-08-19
Separate out "public API" utils from private ones


domenic:
In particular, jsdom at least uses:

- wrapperForImpl
- implForWrapper
- tryWrapperForImpl
- tryImplForWrapper

But it imports these from the same generated utils.js file that also exports a bunch of other stuff, like mixin, implSymbol, iterInteralSymbol, etc.

Instead we should have something like webidl2js-utils.js (public API, those above) and webidl2js-wrapper-helpers.js (private API, the rest). Of course, webidl2js-wrapper-helpers.js can re-export the contents of webidl2js-utils.js so that the wrappers only have to import one file.

PR #68, opened 2017-09-06
Fix mixins, stringifiers, and static attributes

### LINKED ISSUE #49: Rework mixins
The spec [says](https://heycam.github.io/webidl/#es-implements-statements):

> For operations, where the property is a data property with a **Function** object value, each copy of the property must have distinct **Function** objects.

We are not doing that.

### LINKED ISSUE #67: mixin does not properly mix in toString
This currently impacts HTMLAnchorElement. The stringifier from HTMLHyperlinkElementUtils isn't properly mixed into it.

TimothyGu:
This requires a change in jsdom to move idlUtils.mixin to jsdom, as it will no longer be needed or included in webidl2js. I could keep it in webidl2js for the time being however, until the next major bump.

Note, this does not yet fix https://github.com/jsdom/webidl2js/issues/55

Fixes #49.
Fixes #67.

### DIFF
--- a/lib/constructs/interface.js
+++ b/lib/constructs/interface.js
@@ -33,7 +33,9 @@ class Interface {
     this.mixins = [];
 
     this.operations = new Map();
+    this.staticOperations = new Map();
     this.attributes = new Map();
+    this.staticAttributes = new Map();
     this.constants = new Map();
 
     this.indexedGetter = null;
@@ -55,14 +57,17 @@ class Interface {
         continue;
       }
       if (!definingInterface) {
+        let key;
         switch (member.type) {
           case "operation":
-            if (member.name && !this.operations.has(member.name)) {
-              this.operations.set(member.name, new Operation(this.ctx, this, this.idl, member));
+            key = member.static ? "staticOperations" : "operations";
+            if (member.name && !this[key].has(member.name)) {
+              this[key].set(member.name, new Operation(this.ctx, this, this.idl, member));
             }
             break;
           case "attribute":
-            this.attributes.set(member.name, new Attribute(this.ctx, this, this.idl, member));
+            key = member.static ? "staticAttributes" : "attributes";
+            this[key].set(member.name, new Attribute(this.ctx, this, this.idl, member));
             break;
           case "const":
             this.constants.set(member.name, new Constant(this.ctx, this, this.idl, member));
@@ -1145,7 +1150,7 @@ class Interface {
       this.str += `${this.name}.prototype.forEach = ${expr};`;
     }
 
-    for (const member of this.operations.values()) {
+    for (const member of [...this.operations.values(), ...this.staticOperations.values()]) {
       if (!utils.isOnInstance(member.idl, this.idl)) {
         const data = member.generate();
         this.requires.merge(data.requires);
@@ -1160,7 +1165,7 @@ class Interface {
   }
 
   generateAttributes() {
-    for (const member of [...this.attributes.values(), ...this.constants.values()]) {
+    for (const member of [...this.attributes.values(), ...this.staticAttributes.values(), ...this.constants.values()]) {
       if (member instanceof Attribute && utils.isOnInstance(member.idl, this.idl)) {
         continue;
       }
--- a/lib/constructs/interface.js
+++ b/lib/constructs/interface.js
@@ -61,22 +61,26 @@ class Interface {
         switch (member.type) {
           case "operation":
             key = member.static ? "staticOperations" : "operations";
-            if (member.name && !this[key].has(member.name)) {
-              this[key].set(member.name, new Operation(this.ctx, this, this.idl, member));
+            if (member.name) {
+              if (!this[key].has(member.name)) {
+                this[key].set(member.name, new Operation(this.ctx, this, member));
+              } else {
+                this[key].get(member.name).idls.push(member);
+              }
             }
             break;
           case "attribute":
             key = member.static ? "staticAttributes" : "attributes";
-            this[key].set(member.name, new Attribute(this.ctx, this, this.idl, member));
+            this[key].set(member.name, new Attribute(this.ctx, this, member));
             break;
           case "const":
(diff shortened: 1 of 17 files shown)


lib/constructs/interface.js
```diff
@@ -185,7 +221,7 @@ class Interface {
         definingInterface = member[1];
         continue;
       }
-      if (forbiddenMembers.has(member.name)) {
+      if (forbiddenMembers.has(member.name) || member.name && member.name[0] === "_") {
```
domenic: It's worth adding a check that our IDL parsing library properly strips the `_`s from source files before handing them to us. I.e., `attribute boolean _foo` should not trigger an error.

test/cases/Static.idl
```diff
@@ -0,0 +1,6 @@
+interface Static {
+  static attribute DOMString abc;
```
domenic: Nit: I find the alignment here distracting and would prefer no extra spacing.

test/__snapshots__/test.js.snap
```diff
@@ -355,6 +355,357 @@ const Impl = require(\\"../implementations/LegacyArrayClass.js\\");
 "
 `;
 
+exports[`MixedIn.idl 1`] = `
```
domenic: I guess this doesn't directly test that mixing something in to multiple places doesn't reuse the same object... but it does indirectly, as it'd be hard to imagine code generation for the second class that duplicates it.

lib/constructs/interface.js
```diff
@@ -179,16 +231,16 @@ class Interface {
...
+    memberInherited = false;
     for (const member of this.inheritedMembers()) {
-      if (member[0] === inherited) {
-        definingInterface = member[1];
+      if (member === inherited) {
+        memberInherited = true;
         continue;
       }
-      if (forbiddenMembers.has(member.name)) {
+      if (forbiddenMembers.has(member.name) || member.name && member.name[0] === "_") {
```
Sebmaster: Parens this && to reduce (my) confusion about operator precedence.

lib/constructs/interface.js
```diff
@@ -208,6 +260,23 @@ class Interface {
...
+    const iface = this.ctx.interfaces.get(source);
+    if (!iface) {
+      if (this.ctx.options.suppressErrors) {
+        return;
+      }
+      throw new Error(`${source} interface not found (used as a mixin for ${this.name})`);
+    }
+
+    // TODO: are these actually allowed by the spec?
+    if (!this.ctx.options.suppressErrors) {
```
Sebmaster: Remove this conditional. `suppressErrors` should only be used for stuff we know doesn't work but we know is "safe" to ignore.

TimothyGu:
@domenic 

Comments addressed.

> What does "Refactor Overloads operation collection" accomplish? 

It's pure refactoring so that it is closer to the spec's language. No new features added or bugs fixed.

----

@Sebmaster

Comments addressed.

> Do we really have to use this sentinel in the inherited member call? Can we just split this up into two "iterators"?

No, and done.

lib/constructs/interface.js
```diff
@@ -163,9 +122,98 @@ class Interface {
...
+    for (const member of this.members(seen)) {
+      let key;
+      switch (member.type) {
+        case "operation":
+          key = member.static ? "staticOperations" : "operations";
+          if (member.name) {
+            if (!this[key].has(member.name)) {
+              this[key].set(member.name, new Operation(this.ctx, this, member));
+            } else {
+              this[key].get(member.name).idls.push(member);
```
Sebmaster: Is the order of these guaranteed? We're sometimes accessing idls[0]. How do we know that that's the correct one? What's in the others?

TimothyGu: The order is not guaranteed. Here are the specific cases in which `idls[0]` is accessed:

- `utils.isOnInstance(this.idls[0], this.interface.idl)`

   This checks for `[Unforgeable]` extended attribute on `this.idls[0]` and returns true if that is the case. The Web IDL [forbids](https://heycam.github.io/webidl/#Unforgeable) some overloads of an operation having that extended attribute and others of the same operation do not:

   > If it does appear on an operation, then it must appear on all operations with the same identifier on that interface.

   We do not enforce this requirement at this moment, and a TODO is included:

   https://github.com/jsdom/webidl2js/blob/398c67c78193068bd018c19daa271b4c4c073adc/lib/constructs/operation.js#L28

   But assuming the input IDL is well-formed, checking the first overload only should suffice.

- `this.idls[0].name || this.name`

   This is a somewhat hacky way of supporting named stringifiers. Basically in case of

   ```webidl
   stringifier DOMString opName();
               DOMString opName(DOMString input);
   ```

   we create two `Operation`s:

   1. one with `op.name` being `"opName"`, but `op.idls` having both overloads, and
   2. the other's `op.name` property manually overridden with `"toString"`, but `op.idls[0]` is still `"opName"`, and `op.idls.length` is always 1.

   For the first case, since all overloads have the same name, choosing which one doesn't matter. For the second case, since there is always going to be one overload, the first and only one it is.

   OTOH, for a more basic case of 

   ```webidl
   stringifier;
   ```

   we know there is only always going to be one overload.

- `this.idls[0].stringifier`

   See above. This can be true only if there is one item in `this.idls`.

I can add some comments and asserts in the code regarding the last two items if that makes it a bit clearer?

Sebmaster: I'm fine with a single comment explaining this here (or at the access points, whatever seems clearer).

TimothyGu:
Just a note when publishing a new version: this is a breaking change as it removes `mixin()` from the utils file.

domenic:
I've decided that this is not a breaking change because `mixin()` is undocumented. I think that is the right policy going forward. But, it ended up being done as part of v8.0.0 regardless, so we're safe even if some people don't agree.

In practice, what this means is that fixing #56 and #52 are IMO not breaking changes since they change undocumented APIs and don't change how you write an impl class.

On the other hand I counted #72 as breaking since it changes how you would write an impl class.

PR #70, opened 2017-09-16
Add support for enumeration types

### LINKED ISSUE #28: Add enum validation
Includes DOMString conversion and then checking that it's inside a predefined set. Would be pretty nice to not have to code that ourselves :)

TimothyGu:
Fixes #28.

### DIFF
--- a/lib/constructs/enumeration.js
+++ b/lib/constructs/enumeration.js
@@ -10,13 +10,14 @@ class Enumeration {
 
   generate() {
     const values = new Set(this.idl.values);
-    if (!this.ctx.options.suppressErrors && values.size !== this.idl.values.length) {
+    if (values.size !== this.idl.values.length) {
       throw new Error(`Duplicates found in ${this.name}'s enumeration values`);
     }
 
     this.str += `
       const enumerationValues = new Set(${JSON.stringify([...values])});
       module.exports = {
+        enumerationValues,
         convert(value, { context = "The provided value" } = {}) {
           const string = \`\${value}\`;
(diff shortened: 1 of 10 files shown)


lib/constructs/enumeration.js
```diff
@@ -0,0 +1,38 @@
...
+  constructor(ctx, idl) {
+    this.ctx = ctx;
+    this.idl = idl;
+    this.name = idl.name;
+    this.str = null;
+  }
+
+  generate() {
+    const values = new Set(this.idl.values);
+    if (!this.ctx.options.suppressErrors && values.size !== this.idl.values.length) {
```
domenic: Per @Sebmaster it looks like we're not using suppressErrors for this type of thing, only for things that are "unavoidable" like not having every interface type implemented.

test/__snapshots__/test.js.snap
```diff
@@ -80,6 +80,155 @@ module.exports = {
...
+
+    return utils.tryWrapperForImpl(this[impl][\\"attr\\"]);
+  },
+
+  set(V) {
+    if (!this || !module.exports.is(this)) {
+      throw new TypeError(\\"Illegal invocation\\");
+    }
+
+    V = convertRequestDestination(V, { context: \\"Failed to set the 'attr' property on 'Enum': The provided value\\" });
```
domenic: This isn't correct, per https://heycam.github.io/webidl/#dfn-attribute-setter step 3.6

TimothyGu:
Comments addressed.

PR #72, opened 2017-09-20
Fix access to implementation of static operations

Zirro:
I ran into the error `Impl.escape is not a function` while implementing the following IDL:

```webidl
interface CSS {
  static CSSOMString escape(CSSOMString ident);
};
```

This pull request resolves the issue by using `Impl.implementation` rather than `Impl` directly. Given that I am not familiar with all parts of webidl2js, I'm not entirely convinced that this is the correct solution or merely one which obscures a larger issue.

### DIFF
--- a/lib/constructs/attribute.js
+++ b/lib/constructs/attribute.js
@@ -41,8 +41,8 @@ class Attribute {
 
     if (this.static) {
       brandCheck = "";
-      getterBody = `return Impl["${this.idl.name}"];`;
-      setterBody = `Impl["${this.idl.name}"] = V;`;
+      getterBody = `return Impl.implementation["${this.idl.name}"];`;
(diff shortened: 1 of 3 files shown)


Sebmaster:
> But I guess that doesn't really make much sense any more because there's the class syntax.

Was actually in anticipation of introducing more hooks which should not be appended to the impl class directly (something like setup, etc).

Sebmaster:
Yes, we want to change this for attributes as well, LGTM though.

Zirro:
@TimothyGu / @Sebmaster Thanks for the reviews, and the explanation about the pre-ES6 implementation! I'll make the suggested changes tomorrow.

Zirro:
Made the change for attributes as well. This should be ready for merging now.

TimothyGu:
Post-merge LGTM.

Issue #53, opened 2017-08-19
Generated file mixedInto array always seems empty?


domenic:
Browsing through several important mixins in jsdom, it always seems empty. Can we remove it?

domenic:
I guess the intent was to make is/isImpl work, but they don't currently since it's always empty.

TimothyGu:
It looks empty but the array is populated at runtime:

https://github.com/jsdom/webidl2js/blob/fb1698fca5354ccac4fb338c468ca68f1fb463e6/lib/constructs/interface.js#L361

domenic:
Oh, interesting. I think we should add a comment to the output explaining this; we can use this issue to track that. In the meantime I'll fix the readme.

PR #76, opened 2017-09-24
More accurate brand checks

### LINKED ISSUE #62: Brand checks not strict enough
This is low priority unless someone can think of an amazing fix. I'm mostly just documenting it for posterity.

In jsdom, you can fool brand checks by doing things like

```js
const obj = Object.create(document.createElement("p"));
obj.align; // should throw
```

This is due to how we do our `is()` checks.

TimothyGu:
- Check the impl symbol is an own property of the received object.
- Make mixins' `.is()` methods use `iface.is()` instead of `instanceof
  iface.interface`
- Add the interface's `.is()` to `.isMixedInto` of all consequential interfaces.

Fixes #62.

### DIFF
--- a/test/__snapshots__/test.js.snap
+++ b/test/__snapshots__/test.js.snap
@@ -195,14 +195,17 @@ module.exports = {
     return iface;
   }, // createInterface
 
-  mixedInto: [],
+  // When an interface-module that implements this interface as a mixin is loaded, it will append its own \`.is()\`
+  // method into this array. It allows objects that directly implements *those* interfaces to be recognized as
+  // implementing this mixin interface.
+  _mixedIntoPredicates: [],
   is(obj) {
     if (obj) {
-      if (obj[impl] instanceof Impl.implementation) {
+      if (utils.hasOwn(obj, impl) && obj[impl] instanceof Impl.implementation) {
         return true;
       }
-      for (let i = 0; i < module.exports.mixedInto.length; ++i) {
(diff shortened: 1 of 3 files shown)


test/__snapshots__/test.js.snap
```diff
@@ -195,14 +195,14 @@ module.exports = {
     return iface;
   }, // createInterface
 
-  mixedInto: [],
+  isMixedInto: [],
   is(obj) {
     if (obj) {
-      if (obj[impl] instanceof Impl.implementation) {
+      if (Object.prototype.hasOwnProperty.call(obj, impl) && obj[impl] instanceof Impl.implementation) {
```
domenic: Should we put a `hasOwn(x, y)` in utils.js maybe?

test/__snapshots__/test.js.snap
```diff
@@ -195,14 +195,14 @@ module.exports = {
     return iface;
   }, // createInterface
 
-  mixedInto: [],
+  isMixedInto: [],
```
domenic: Using `is` prefixes for non-boolean variables (an array in this case) is especially confusing.

Maybe this should have an underscore prefix since it appears to be an implementation detail?

So `_mixedIntoPredicates` perhaps?

domenic:
Maybe also fix https://github.com/jsdom/webidl2js/issues/53 while here?

TimothyGu:
Comments addressed.

PR #77, opened 2017-09-26
Fix typo: new.target => this.target

aristov:


aristov:
Sorry, I've never seen such syntax before.
Live and learn!

PR #83, opened 2017-10-06
Work around a webkit particularity

### LINKED ISSUE #78: Work around Safari/webkit `new.target can't come after a prefix operator` issue
Safaru currently fails on prototypes created using webidl2js with 

> new.target can't come after a prefix operator

This is a silly error and entirely Safari's fault. Literally you're not allowed to (from `whatwg-url`:

    function URL(url) {
      if (!new.target) {
        throw new TypeError(

Because [the bang throws webkit](https://stackoverflow.com/questions/45578781/new-target-with-a-prefix-operator). [Webkit has now been fixed](https://bugs.webkit.org/show_bug.cgi?id=157970#c17), but the current Safari in macOS High Sierra and iOS 11.02 `13604.1.38.1.6` still has the error.

    if (new.target === undefined) {

works fine as a workaround. Ideally Apple would already have a new Safari rolled out, but in the meantime, and for older devices, can we get `webidl2js` to use the workaround?

Sebmaster:
A version of webkit doesn't allow new.target after a prefix operator.

Fixes #78.

### DIFF
--- a/lib/constructs/interface.js
+++ b/lib/constructs/interface.js
@@ -332,7 +332,7 @@ class Interface {
       if (minConstructor.nameList.length !== 0) {
         const plural = minConstructor.nameList.length > 1 ? "s" : "";
         this.str += `
-          if (!new.target) {
+          if (new.target === undefined) {
             throw new TypeError("Failed to construct '${this.name}'. Please use the 'new' operator; this constructor " +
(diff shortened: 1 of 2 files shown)


mikemaccana:
Thanks Deb and Dom! 😃

Issue #80, opened 2017-10-06
transformer.addSource(idl, impl) should interpret both paths as relative to the same base directory


danyao:
When using relative paths with `transformer.addSource(idl, impl)`, `idl` is interpreted relative to the base directory, where as `impl` is interpreted relative to the output directory of `transformer.generate`.

For example, if my Node.js looks like this:
```javascript
transformer.addSource("idl", "impl");  // [*]
transformer.generate("wrappers");
```
And I have the following directory structure:
```
./
  idl/
    foo.webidl
  impl/
    Foo-impl.js
  wrappers/
```
The generated file `./wrappers/Foo.js` will contain the following that can't be resolved:
```javascript
const Impl = require("./impls/Foo-impl.js");
```

Changing Node.js to the following works, but looks unintuitive:
```javascript
transformer.addSource("idl", "../impl");
```

TimothyGu:
We can `path.resolve()` the paths before using them in `addSource()` and `generate()`. @danyao can you submit a PR to fix that?

Sebmaster:
Breaking change btw, need to major for it.

danyao:
Yep I'll take a stab.

danyao:
How do I "major for it"?

TimothyGu:
@danyao Bumping version is part of the process of making releases. We'll take care of it. (I think @Sebmaster said that as a reminder.)

PR #85, opened 2017-10-11
Track IDL and Impl directories using absolute paths in transformer.js

danyao:
Convert IDL and Impl directories to absolute paths before using them in addSource() and generate() as a fix for #80.

Ran npm test and everything passes.

### DIFF
--- a/lib/transformer.js
+++ b/lib/transformer.js
@@ -22,7 +22,7 @@ class Transformer {
       }
     });
 
-    this.sources = [];  // Absolute paths to the IDL and Impl directories.
+    this.sources = []; // Absolute paths to the IDL and Impl directories.
     this.utilPath = null;
   }
 
@@ -178,8 +178,7 @@ class Transformer {
       let source = obj.toString();
 
       const implDir = obj.opts.implDir.replace(/\\/g, "/"); // fix windows file paths
-      let implFile = path.relative(
-        outputDir, path.resolve(implDir, obj.name + this.ctx.implSuffix));
+      const implFile = path.relative(outputDir, path.resolve(implDir, obj.name + this.ctx.implSuffix));
 
       let relativeUtils = path.relative(outputDir, this.utilPath).replace(/\\/g, "/");
       if (relativeUtils[0] !== ".") {
--- a/lib/transformer.js
+++ b/lib/transformer.js
@@ -177,8 +177,8 @@ class Transformer {
     for (const obj of interfaces.values()) {
       let source = obj.toString();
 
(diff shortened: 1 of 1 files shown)


lib/transformer.js
```diff
@@ -177,11 +177,9 @@ class Transformer {
     for (const obj of interfaces.values()) {
       let source = obj.toString();
 
-      const implDir = path.relative(outputDir, obj.opts.implDir).replace(/\\/g, "/"); // fix windows file paths
-      let implFile = implDir + "/" + obj.name + this.ctx.implSuffix;
-      if (implFile[0] !== ".") {
-        implFile = "./" + implFile;
```
TimothyGu: Is this bit not necessary anymore?

lib/transformer.js
```diff
@@ -177,11 +177,9 @@ class Transformer {
...
-      }
+      const implDir = obj.opts.implDir.replace(/\\/g, "/"); // fix windows file paths
+      let implFile = path.relative(
+        outputDir, path.resolve(implDir, obj.name + this.ctx.implSuffix));
```
TimothyGu: We wrap at 120 cols, so these lines could be coalesced.

danyao: Done.

lib/transformer.js
danyao: Thanks for catching it. It turns out path.relative() does not prepend the necessary "./" (contrary to my impression). I only tested with an impl path that's not directly under the output directory so didn't catch this edge case. Restored the "./" fix.

lib/transformer.js
```diff
@@ -177,8 +177,8 @@ class Transformer {
     for (const obj of interfaces.values()) {
       let source = obj.toString();
 
-      const implDir = path.relative(outputDir, obj.opts.implDir).replace(/\\/g, "/"); // fix windows file paths
-      let implFile = implDir + "/" + obj.name + this.ctx.implSuffix;
+      const implDir = obj.opts.implDir.replace(/\\/g, "/"); // fix windows file paths
+      let implFile = path.relative(outputDir, path.resolve(implDir, obj.name + this.ctx.implSuffix));
```
Sebmaster: If you're calling relative/resolve here, I think you're reintroducing windows slashes you'll need to correct.

danyao: Ah I see. Moved Windows path fix to last.

danyao:
Newbie question: should I close the pull request?

domenic:
No :). We'll merge it. Usually we process this sort of thing on weekends, since this is a spare-time project for all of us, but I can merge it now.

danyao:
Got it. I didn't mean to rush you. Thanks everyone for the guidance!

PR #86, opened 2017-10-15
 Fully implement overloaded and variadic ops

TimothyGu:


domenic:
Wow. I looked through things, especially the snapshot output, and it all looks in order. It does kind of bloat our code size a bit (even for non-overloads), with the curArg-convert-push sequence, but it's fine, and if we wanted to optimize the output in the future we could.

I still think dispatching to differently-named impl overloads would be nicer in most cases, but am happy to wait to see how things play out in jsdom before we go that route. This takes care of the hard part.

I guess this isn't even a breaking change, really.

Issue #87, opened 2017-10-28
Implement [HTMLConstructor] extended attribute


mraerino:
*Follow-up to #44*

**Spec: [[HTMLConstructor]](https://html.spec.whatwg.org/multipage/dom.html#htmlconstructor)**

For the sake of https://github.com/tmpvar/jsdom/issues/1030 i really want to make the `[HTMLConstructor]` extended attribute work.

Are there any plans to be able to implement extended attributes from `jsdom`? How would this work?

The way I see this, `[HTMLConstructor]` could behave just like `[Constructor]` with some specials implemented:

- Check: No `[Constructor]` or `[NoInterfaceObject]` attributes present
- Check: Needs to be inheriting from `HTMLElement`
- Passes `new.target` to the implementation constructor as `newTarget`

I would like to get this in here first, so when this is through we can start using `[HTMLConstructor` in jsdom.

I'm happy to implement, but would need some proper guidance on the best way.

Sebmaster:
We don't usually check the correctness of input idls, so that's not really necessary.

Should just be able to save this in `privateData` and pass it to the impl constructor [here](https://github.com/jsdom/webidl2js/blob/e8604c070d8f69784e0d54399bf0ddf44ad800ed/lib/constructs/interface.js#L349).

mraerino:
Closing in favor of implementation in #88

PR #88, opened 2017-10-28
Implement [HTMLConstructor] extended attribute

mraerino:
Implemented per feedback in #87

**Spec: [[HTMLConstructor]](https://html.spec.whatwg.org/multipage/dom.html#htmlconstructor)**

### DIFF
--- a/test/__snapshots__/test.js.snap
+++ b/test/__snapshots__/test.js.snap
@@ -523,6 +523,173 @@ const Impl = require(\\"../implementations/Factory.js\\");
 "
 `;
 
+exports[`HTMLConstructor.webidl 1`] = `
+"\\"use strict\\";
+
+const conversions = require(\\"webidl-conversions\\");
+const utils = require(\\"./utils.js\\");
+
+const impl = utils.implSymbol;
+
+function HTMLConstructor() {
+  iface.setup(this, [], { newTarget: new.target });
+}
+
+Object.defineProperty(HTMLConstructor, \\"prototype\\", {
+  value: HTMLConstructor.prototype,
+  writable: false,
+  enumerable: false,
+  configurable: false
+});
+
+HTMLConstructor.prototype.click = function click() {
+  if (!this || !module.exports.is(this)) {
+    throw new TypeError(\\"Illegal invocation\\");
+  }
+
+  return this[impl].click();
+};
+
+HTMLConstructor.prototype.focus = function focus() {
+  if (!this || !module.exports.is(this)) {
+    throw new TypeError(\\"Illegal invocation\\");
+  }
+
+  const args = [];
+  for (let i = 0; i < arguments.length && i < 1; ++i) {
+    args[i] = arguments[i];
+  }
+
+  if (args[0] !== undefined) {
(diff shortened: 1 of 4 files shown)


test/__snapshots__/test.js.snap
```diff
@@ -395,6 +395,173 @@ const Impl = require(\\"../implementations/Factory.js\\");
...
+exports[`HTMLConstructor.webidl 1`] = `
+"\\"use strict\\";
+
+const conversions = require(\\"webidl-conversions\\");
+const utils = require(\\"./utils.js\\");
+
+const impl = utils.implSymbol;
+
+function HTMLConstructor() {
+  iface.setup(this, [], { newTarget: new.target });
```
TimothyGu: The constructor could be improved by:

1. implementing the following step in the spec:

   > If NewTarget is equal to the active function object, then throw a TypeError and abort these steps.

2. throwing in case `new.target` is undefined

   > When invoked with [[Call]], the NewTarget value is undefined, and so the algorithm below will immediately throw. 

I.e.

```js
if (new.target === undefined) {
  throw new TypeError("Failed to construct 'HTMLConstructor'. Please use the 'new' operator; this constructor " +
                      "cannot be called as a function.");
}
if (new.target === HTMLConstructor) {
  throw new TypeError("Illegal constructor");
}
```

domenic:
I don't understand how this implements the spec at all. For example it doesn't even contain step 1 of the spec. @Sebmaster, why did you think this would work?

I don't believe it's possible to implement the spec in jsdom's current architecture, as discussed on various custom elements issues on jsdom.

Sebmaster:
@domenic Sorry, didn't actually check the spec for this, just answered the question in the initial post.

TimothyGu:
@domenic I don't see any reason not to have an escape hatch for [`HTMLConstructor`], that depends on the implementation file to provide full spec behaviors. As for step 1 of the definition of [`HTMLConstructor`] in the spec, it really should be step 2, as it has no side effects and declares a variable that will only be used in step 3.

mraerino:
@domenic do you suggest implementing Step 2 from the spec in `webidl2js`?
I rather thought to leave any algorithm implementation of the spec to `jsdom` in order to have a minimal footprint here.

Another way would be to provide a pluggable architecture which will allow jsdom to provide functionality for extended attributes.

What do you think?

mraerino:
Ok, I noticed it makes way more sense to implement the check in step 2 inside the idl constructor. Let me get that right.

domenic:
@TimothyGu the entire point of HTMLConstructor is to abstract that behavior into the IDL-generating code layer. I don't think we should leave it to impl files.

> Another way would be to provide a pluggable architecture which will allow jsdom to provide functionality for extended attributes.

Yeah, that would also work.

mraerino:
I rebased upon `master` and added the check suggested by @TimothyGu
Seems better?

domenic:
I'm still not interested in accepting this given that it doesn't implement the spec, sorry. Step 2 was just an example; you'd need to implement steps 1-13. And, as discussed elsewhere, step 1 is not possible in jsdom's current architecture.

PR #91, opened 2017-11-12
Update dependencies

TimothyGu:
webidl2.js has dropped official support for Node.js v6.x in v8.0.0 (which we still want to support), but it still fully works with it (as shown by our test suite). Let's wait until https://github.com/w3c/webidl2.js/pull/102#issuecomment-343706702 is resolved before this is merged.

### DIFF
--- a/lib/constructs/constant.js
+++ b/lib/constructs/constant.js
@@ -14,11 +14,11 @@ class Constant {
   generate() {
     const body = `
       Object.defineProperty(${this.interface.name}, "${this.idl.name}", {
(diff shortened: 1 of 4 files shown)


domenic:
FWIW jsdom doesn't support all LTS by policy. If there's something in 7, 8, or even 9 that makes our lives easier, we'd change the required Node version, as we have in the past. I just haven't seen such a thing yet... Although the VM module fixes certainly help with a lot of bugs.

TimothyGu:
webidl2.js has added back support for Node.js v6.x. Let's merge this.

PR #95, opened 2017-11-25
[Unscopeable] is now [Unscopable]

Zirro:
This changed all the way back in https://github.com/heycam/webidl/issues/103.

It is a breaking change, though I doubt many will be affected.


Zirro:
That update is part of https://github.com/tmpvar/jsdom/pull/2053.

PR #98, opened 2017-12-17
Use class/object method definitions for operations 

### LINKED ISSUE #97: "prototype" property exists for operations
```js
> const { window } = new (require('jsdom').JSDOM)();
> typeof window.Document.prototype.querySelector.prototype
'object'
```

compared with in a browser:

```js
> typeof window.Document.prototype.querySelector.prototype
"undefined"
```

TimothyGu:
Fixes #97.

### DIFF
--- a/test/__snapshots__/test.js.snap
+++ b/test/__snapshots__/test.js.snap
@@ -582,91 +582,88 @@ const iface = {
   _internalSetup(obj) {
     Object.defineProperties(
       obj,
-      Object.getOwnPropertyDescriptors(
-        Object.defineProperties(
-          {
-            op() {
-              if (!this || !module.exports.is(this)) {
-                throw new TypeError(\\"Illegal invocation\\");
-              }
-
-              return this[impl].op();
-            },
-            unforgeableOp() {
-              if (!this || !module.exports.is(this)) {
-                throw new TypeError(\\"Illegal invocation\\");
-              }
-
-              return this[impl].unforgeableOp();
-            },
-            get attr() {
-              if (!this || !module.exports.is(this)) {
-                throw new TypeError(\\"Illegal invocation\\");
-              }
-
-              return obj[impl][\\"attr\\"];
-            },
-            set attr(V) {
-              if (!this || !module.exports.is(this)) {
-                throw new TypeError(\\"Illegal invocation\\");
-              }
-
-              V = conversions[\\"DOMString\\"](V, {
-                context: \\"Failed to set the 'attr' property on 'Global': The provided value\\"
-              });
-
-              obj[impl][\\"attr\\"] = V;
-            },
-            get unforgeableAttr() {
-              if (!this || !module.exports.is(this)) {
-                throw new TypeError(\\"Illegal invocation\\");
-              }
-
-              return obj[impl][\\"unforgeableAttr\\"];
-            },
-            set unforgeableAttr(V) {
-              if (!this || !module.exports.is(this)) {
-                throw new TypeError(\\"Illegal invocation\\");
-              }
-
-              V = conversions[\\"DOMString\\"](V, {
-                context: \\"Failed to set the 'unforgeableAttr' property on 'Global': The provided value\\"
-              });
-
-              obj[impl][\\"unforgeableAttr\\"] = V;
-            },
-            get length() {
-              if (!this || !module.exports.is(this)) {
-                throw new TypeError(\\"Illegal invocation\\");
-              }
-
-              return obj[impl][\\"length\\"];
-            },
-            set length(V) {
-              if (!this || !module.exports.is(this)) {
-                throw new TypeError(\\"Illegal invocation\\");
-              }
-
-              V = conversions[\\"unsigned long\\"](V, {
-                context: \\"Failed to set the 'length' property on 'Global': The provided value\\"
-              });
-
-              obj[impl][\\"length\\"] = V;
-            },
-            [Symbol.iterator]: Array.prototype[Symbol.iterator],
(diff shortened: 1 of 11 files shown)


test/__snapshots__/test.js.snap
```diff
@@ -523,47 +545,255 @@ const Impl = require(\\"../implementations/Factory.js\\");
...
+  },
+  createImpl(constructorArgs, privateData) {
+    let obj = Object.create(Global.prototype);
+    obj = this.setup(obj, constructorArgs, privateData);
+    return utils.implForWrapper(obj);
+  },
+  _internalSetup(obj) {
+    Object.defineProperties(
+      obj,
+      Object.getOwnPropertyDescriptors(
```
domenic: This indirection seems unfortunate; any way to avoid it?

test/__snapshots__/test.js.snap
```diff
@@ -523,47 +545,255 @@ const Impl = require(\\"../implementations/Factory.js\\");
...
       throw new TypeError(\\"Illegal invocation\\");
     }
 
     return this[impl][\\"length\\"];
+  }
+}
+Object.setPrototypeOf(LegacyArrayClass.prototype, Array.prototype);
+Object.defineProperties(LegacyArrayClass.prototype, {
+  length: {
+    configurable: true,
```
domenic: configurable is true by default; no need to set it explicitly, I don't think

test/__snapshots__/test.js.snap
```diff
@@ -412,28 +418,44 @@ module.exports = {
...
+      }
     });
-
-    Object.defineProperty(Factory.prototype, Symbol.toStringTag, {
-      value: \\"Factory\\",
-      writable: false,
-      enumerable: false,
-      configurable: true
+    Object.defineProperties(Factory, {
+      prototype: {
```
domenic: Class declarations by default generate a prototype with this descriptor, so this is not necessary.

test/__snapshots__/test.js.snap
TimothyGu: I could do `defineProperties(getOwnPropertyDescriptors()); defineProperties();` rather than `defineProperties(getOwnPropertyDescriptors(defineProperties()));` as it is currently, but there aren't many ways of avoiding the `getOwnPropertyDescriptors()` because I want to use the method syntax to reduce even more `defineProperties()`.

test/__snapshots__/test.js.snap
TimothyGu: Interesting, I was not aware of that.

test/__snapshots__/test.js.snap
TimothyGu: > in general I'd love to reduce the redundant property descriptor values.

I could work on this, but it just seems to be better form to have everything explicitly written out rather than depending some flags on default behaviors while modifying others.

test/__snapshots__/test.js.snap
domenic: Maybe Object.assign + defineProperties?

test/__snapshots__/test.js.snap
domenic: I just get a bit worried about code size, but maybe it's fine... I like the succinctness of only changing what's actually different, I guess. Shrug.

test/__snapshots__/test.js.snap
TimothyGu: `Object.assign` will kill accessors, and accessors can't be specified through `defineProperties` due to the requirement that `Object.getOwnPropertyDescriptor(Interface.prototype, "accessor").get.name === "get accessor"` (of course, without another layer of `defineProperty("name")`).

test/__snapshots__/test.js.snap
TimothyGu: You're right... the delta-only output does look very nice. PR updated.

test/cases/Global.webidl
```diff
@@ -1,4 +1,4 @@
-[Global]
+[Global=Global]
```
domenic: Should we throw an error if the RHS is absent?

domenic:
Hmm we should also update the README excerpt at some point, although I think it was out of date already.

test/cases/Global.webidl
TimothyGu: We don't really claim to support `[Global]` in the README, but sure.

domenic: Hmm yeah, I'm a little confused what the state of our support is there. (For [Global], not necessarily [Global] + named properties.)

TimothyGu:
> accessors use Object.defineProperties with the full descriptor (separate get/set functions)

Again, if we do that then we'd need additional `Object.defineProperty`'s to fix up the `name` of those getters/setters (per [spec](https://heycam.github.io/webidl/#dfn-attribute-getter)), in the following fashion

```js
Object.defineProperties(obj, {
  prop: {
    get() {
      // ...
    },
    set(a) {
      // ...
    },
    // ...
  }
});
Object.defineProperty(Object.getOwnPropertyDescriptor(obj, "prop").get, "name", { value: "get prop" });
Object.defineProperty(Object.getOwnPropertyDescriptor(obj, "prop").set, "name", { value: "set prop" });
```

which isn't much of a gain at all.

> we should also update the README excerpt at some point

Will do.

domenic:
Hah wow, that really just drives home how poor the choices are in ES for this kind of stuff. Ah well. Thanks for walking me through the reasoning.

TimothyGu:
Merged in de5b817a0c378ad572d5333748831c776a64af6c...b26285a2fee18f2989874c56f55d94e52c977781.

PR #100, opened 2017-12-25
Remove unnecessary usage of suppressErrors

TimothyGu:
`suppressErrors` is only intended to be used to suppress errors related to missing interface definitions, not a get-out-of-jail-free card for all situations.


domenic:
I wonder how much jsdom needs suppressErrors these days anyway?

TimothyGu:
@domenic Not much, only for certain things we still don't support like callback interfaces and callback functions.

Issue #59, opened 2017-08-19
Can implementation classes default-export implementation, instead of having a single export?


domenic:
We don't have any potential exports beside `implementation`, right? This would be a decent amount of churn on the jsdom side, but it seems easier to explain.

Sebmaster:
The original plan here was to allow for stuff like impl setup methods which are not part of the exposed interface. Not sure if we still need that for anything...

domenic:
https://github.com/jsdom/webidl2js/issues/58#issuecomment-323542251 has some discussion of why this might be a good idea, although it's not concrete enough for me to say that we definitely want to keep the current form, so I'll leave this open for now.

ExE-Boss:
I believe that the following:
```js
export default class SomeInterface {
	//
}
```
is better than:
```js
class SomeInterface {
	//
}
export { SomeInterface as implementation };
```

ExE-Boss:
I’m doing this in #141.

PR #101, opened 2017-12-25
Add a way to expose interfaces automatically

TimothyGu:
jsdom PR coming up!


ExE-Boss:
This is&nbsp;completely unrelated to&nbsp;#59, which is&nbsp;about changing the&nbsp;name of&nbsp;the&nbsp;export in&nbsp;implementation&nbsp;files.

PR #102, opened 2017-12-26
Provide polyfill for Object.getOwnPropertyDescriptors

TimothyGu:
We still support Node.js v6.x, which does not have `Object.getOwnPropertyDescriptors()`.

### DIFF
--- a/lib/output/utils.js
+++ b/lib/output/utils.js
@@ -9,6 +9,30 @@ function hasOwn(obj, prop) {
   return Object.prototype.hasOwnProperty.call(obj, prop);
 }
 
+const getOwnPropertyDescriptors = typeof Object.getOwnPropertyDescriptors === "function" ?
+  Object.getOwnPropertyDescriptors :
+  // Polyfill exists before we require Node.js v6.x
+  // https://tc39.github.io/ecma262/#sec-object.getownpropertydescriptors
+  obj => {
+    if (obj === undefined || obj === null) {
+      throw new TypeError("Cannot convert undefined or null to object");
+    }
+    const ownKeys = Reflect.ownKeys(obj);
+    const descriptors = {};
+    for (const key of ownKeys) {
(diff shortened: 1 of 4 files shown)


lib/output/utils.js
```diff
@@ -9,6 +9,30 @@ function hasOwn(obj, prop) {
   return Object.prototype.hasOwnProperty.call(obj, prop);
 }
 
+const getOwnPropertyDescriptors = typeof Object.getOwnPropertyDescriptors === "function" ?
+  Object.getOwnPropertyDescriptors :
+  // Polyfill exists before we require Node.js v6.x
```
Zirro: Either the wording or the Node.js-version seems a bit confusing here - could this say "Polyfill needed until we require Node.js v8.x" instead?

Zirro:
I doubt we'll manage to support Node.js v6 until the official maintenance ends in April 2019 given all the `vm`-bugs, but this looks good for now :-)

LGTM with small nit (from looking at the code rather than running it)

lib/output/utils.js
TimothyGu: Not sure what I was thinking… that sounds perfect.

lib/output/utils.js
```diff
@@ -9,6 +9,30 @@ function hasOwn(obj, prop) {
...
+  // https://tc39.github.io/ecma262/#sec-object.getownpropertydescriptors
+  obj => {
+    if (obj === undefined || obj === null) {
+      throw new TypeError("Cannot convert undefined or null to object");
+    }
```
TimothyGu: This seems to be missing a `obj = Object(obj)`… will fix.

TimothyGu: Fixed.

domenic:
I might not call this a "polyfill" in the commit message since it doesn't actually add a property to Object. E.g. if I saw "now includes a polyfill for O.gOPDs" in the changelog I'd be worried that one of my dependencies had started modifying the global objects.

Issue #104, opened 2018-03-07
Support interface mixins


TimothyGu:
https://heycam.github.io/webidl/#interface-mixin

https://heycam.github.io/webidl/#includes-statement

PR #105, opened 2018-03-28
Support interface mixins and includes statements

TimothyGu:
Fixes: #104


domenic:
Can we remove support for "implements" at the same time?

PR #108, opened 2018-06-22
Wrap generic return values

qwtel:
Fixes wrapping of implementation classes for `Promise`, `sequence`, and `record` return types.

Currently, when a operation returns a promise, array, or object, the generated output will be

```js
return utils.tryWrapperForImpl(${callOn}.${implFunc}(${argsSpread}))
```

which does not accomplish the goal of wrapping the implementation class for the objects contained inside.

Using promises as an example, the output should instead be:

```js
const res = ${callOn}.${implFunc}(${argsSpread});
return res && res.then ? res.then(utils.tryWrapperForImpl) : res;
```

### DIFF
--- a/test/__snapshots__/test.js.snap
+++ b/test/__snapshots__/test.js.snap
@@ -4817,7 +4817,8 @@ class URLSearchParams {
       });
       args.push(curArg);
     }
-    return utils.tryWrapperForImpl(this[impl].getAll(...args));
+    const res = this[impl].getAll(...args);
+    return res && res.map ? res.map(utils.tryWrapperForImpl) : res;
   }
 
   has(name) {


lib/constructs/operation.js
```diff
@@ -79,14 +80,46 @@ class Operation {
...
       `;
     } else {
-      str += `
-        return utils.tryWrapperForImpl(${callOn}.${implFunc}(${argsSpread}));
-      `;
+      switch (this.generic) {
+        case "Promise": {
+          str += `
+            const res = ${callOn}.${implFunc}(${argsSpread});
+            return res && res.then ? res.then(utils.tryWrapperForImpl) : res;
```
TimothyGu: Let's make the code stricter. Only check `res` is null if the return type is nullable, and if it's not null (or the return type is not nullable) just do `res.then(utils.tryWrapperForImpl)`.

Ditto for the two blocks below.

lib/constructs/operation.js
```diff
@@ -13,6 +13,7 @@ class Operation {
     this.idls = [idl];
     this.name = idl.name;
     this.static = idl.static;
+    this.generic = idl.idlType.generic;
```
TimothyGu: This assumes that the return type of all potential overloads are the same. (That's why `this.idls` is an array – all overloaded operations with the same name are grouped into the same `Operation` object.)

domenic:
Resurrecting this would be helpful for the Streams Standard reference implementation, which has a `tee()` method that returns an array. Currently I need to manually map the results using `wrapperForImpl`.

PR #116, opened 2018-12-06
chore(package): update webidl2

saschanaz:
~~Found that 10.3.0 doesn't break JSDOM while 10.3.1+ does.~~

Fixed the breakage faced in jsdom/jsdom#2443 so it's now safe to update to 10.3.1+.


domenic:
Thank you so much, @saschanaz!! It is really appreciated for you to do this level of sleuthing, and also follow-up work, on the issue.

PR #120, opened 2019-03-30
Only reflect attributes without a namespace

Zirro:
Following the discussion in https://github.com/jsdom/jsdom/pull/2546. The change has been tested with jsdom.

The `getAttribute()`-method includes a step that lowercases the attribute name while `getAttributeNS()` does not, so it is now done explicitly here. This is necessary for all tests in jsdom to pass.

### DIFF
--- a/lib/constructs/attribute.js
+++ b/lib/constructs/attribute.js
@@ -51,8 +51,8 @@ class Attribute {
         throw new Error("Unknown reflector type: " + this.idl.idlType.idlType);
       }
       const attrName = shouldReflect.rhs && shouldReflect.rhs.value.replace(/_/g, "-") || this.idl.name;
-      getterBody = reflector[this.idl.idlType.idlType].get(objName, attrName);
-      setterBody = reflector[this.idl.idlType.idlType].set(objName, attrName);
+      getterBody = reflector[this.idl.idlType.idlType].get(objName, attrName.toLowerCase());
+      setterBody = reflector[this.idl.idlType.idlType].set(objName, attrName.toLowerCase());
     }
 
     if (utils.getExtAttr(this.idl.extAttrs, "LenientThis")) {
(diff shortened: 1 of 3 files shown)


Sebmaster:
`setAttribute` will always set with a null namespace instead of loosely matching to a pre-existing attribute? Looking at the jsdom source code, we might have a problem there as well.

Zirro:
> setAttribute will always set with a null namespace instead of loosely matching to a pre-existing attribute?

Hmm, probably not. I'll amend this to make the change to `setAttribute()` as well.

> Looking at the jsdom source code, we might have a problem there as well.

https://github.com/jsdom/jsdom/pull/2547

Sebmaster:
> jsdom/jsdom#2547

Was referring to the implementation of setAttribute specifically. But I think that PR has the same problem now. I.e. `draggable` setter will modify `ns:draggable` (if it exists) instead of creating `draggable`.

Zirro:
Indeed, will fix in both PRs soon. The case of manipulating an existing attribute (rather than creating a new one) through `setAttribute()` did not occur to me while I was working on the code last night, presumably as I was rather tired.

lib/constructs/attribute.js
```diff
@@ -51,8 +51,8 @@ class Attribute {
         throw new Error("Unknown reflector type: " + this.idl.idlType.idlType);
       }
       const attrName = shouldReflect.rhs && shouldReflect.rhs.value.replace(/_/g, "-") || this.idl.name;
-      getterBody = reflector[this.idl.idlType.idlType].get(objName, attrName);
-      setterBody = reflector[this.idl.idlType.idlType].set(objName, attrName);
+      getterBody = reflector[this.idl.idlType.idlType].get(objName, attrName.toLowerCase());
+      setterBody = reflector[this.idl.idlType.idlType].set(objName, attrName.toLowerCase());
```
domenic: This really should do only ASCII lowercase, although I guess in our case it'll be fine since no IDL attributes are non-ASCII.

(A refactoring project I've meant to do for a while is publish an ASCII string manipulation package with the various utils we've been re-creating in lots of jsdom projects.)

PR #121, opened 2019-04-20
Fix operation ordering of record type coercions

Sebmaster:
The WebIDL spec type-converts the key before retrieving the value. This is observable when providing a Proxy to this conversion. In that case [[GetOwnProperty]] should be called, but [[Get]] should not be observable.

By doing the key conversion before retrieving the value, we re-throw an error earlier and should be fine.

### DIFF
--- a/lib/types.js
+++ b/lib/types.js
@@ -315,8 +315,9 @@ function generateTypeConversion(ctx, name, idlType, argAttrs = [], parentName, e
           const desc = Object.getOwnPropertyDescriptor(${name}, key);
           if (desc && desc.enumerable) {
             let typedKey = key;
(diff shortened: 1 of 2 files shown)


domenic:
I assume this will be well-tested by some web platform test for Headers?

Sebmaster:
> I assume this will be well-tested by some web platform test for Headers?

Exactly. There's another bug with proxies somewhere here (where we should just throw immediately, without triggering any traps). But this better matches spec for now and I'll dig what the other issue is.

Sebmaster:
> But this better matches spec for now and I'll dig what the other issue is.

Other issue was the ownKeys thing (and a leftover console.log). Should be good to go.

Issue #71, opened 2017-09-16
Factor out IDL-to-ECMAScript type conversion


TimothyGu:
We currently have an incomplete version of the actual algorithms:

https://github.com/jsdom/webidl2js/blob/bf9954ddc186e4d733957aba0bdcde35366fb548/lib/constructs/operation.js#L88-L98

Would be nice to factor it out, for e.g. callback function support.

qwtel:
I've noticed that this section was not doing the right thing for `Promise`, `sequence`, and `record`, so I've made some changes: https://github.com/platformparity/webidl2js/compare/master...platformparity:genericwrap

I didn't factor it out though, nor am I aware that there's an actual algorithm defined somewhere (is this part of WebIDL?).

PR #123, opened 2019-05-27
Add support for callback functions

TimothyGu:


test/__snapshots__/test.js.snap
```diff
@@ -4558,6 +4558,129 @@ const Impl = require(\\"../implementations/URL.js\\");
...
+            \\"convertedFunc.construct()\\"
+        );
+      }
+
+      let returned;
+
+      for (let i = 0; i < args.length; i++) {
+        args[i] = utils.tryWrapperForImpl(args[i]);
+      }
+      if (args.length < 2) {
```
domenic: Reusing the args array is a bit confusing, hmm. I guess it's a bit more efficient though, so I don't know.

Also, I think this code could be simplified as follows:

```js
for (let i = args.length; i < 2; ++i) {
  args[i] = undefined;
}
args.length = 2;
```

And I mean, ideally the generated code would look something like

```js
const convertedArgs = [
  0 in args ? utils.tryWrapperForImpl(args[0]) : undefined,
  1 in args ? utils.tryWrapperForImpl(args[1]) : undefined
];
```

but that level of tuning might not be worth while, I dunno.

test/__snapshots__/test.js.snap
```diff
@@ -4558,6 +4558,129 @@ const Impl = require(\\"../implementations/URL.js\\");
...
+        for (let i = args.length; i < 2; i++) {
+          args[i] = undefined;
+        }
+      } else if (args.length > 2) {
+        args.length = 2;
+      }
+
+      returned = Reflect.apply(value, this, args);
+
+      returned = convertURL(returned, { context: context });
```
domenic: I'm sure I'm just forgetting how things work, but where is convertURL defined?

test/__snapshots__/test.js.snap
```diff
@@ -4558,6 +4558,129 @@ const Impl = require(\\"../implementations/URL.js\\");
...
+
+      return returned;
+    }
+    wrapper.construct = function(...args) {
+      for (let i = 0; i < args.length; i++) {
```
domenic: Should the coercion be shared, e.g. as a module-level function?

test/__snapshots__/test.js.snap
domenic: Can the generated code just use `{ context }`?

test/__snapshots__/test.js.snap
TimothyGu: This (admitted not great) algorithm came from https://heycam.github.io/webidl/#web-idl-arguments-list-converting, which has something that looks like this. I wouldn't be opposed to something that looks like your second option above though, if the complexity is manageable.

test/__snapshots__/test.js.snap
TimothyGu: Hmm weird, looks like a bug.

TimothyGu: We generate this object literal through a call to `Types.generateTypeConversion()`, which takes `"context"` as an argument. We could add a special case, but ideally this would be the job of Prettier, so I'm not too inclined to do that.

test/__snapshots__/test.js.snap
TimothyGu: I'll investigate.

TimothyGu:
I should also add that I haven't yet gotten a chance to test this out on jsdom, as the `convertURL` bug shows. I'll make sure to do that before merging this.

test/__snapshots__/test.js.snap
domenic: I guess it could also be in utils.js, so that the call site would then just look like `convertArgsList(inputArgs, 2)`.

TimothyGu: Eventually there might be more complicated conversion from IDL values to ES values (#71). https://heycam.github.io/webidl/#web-idl-arguments-list-converting takes a Web IDL arguments list, so I'd like to keep it that way.

TimothyGu: Oh wait forget about that. It's not actually dependent on the types.

lib/constructs/callback-function.js
```diff
@@ -0,0 +1,102 @@
...
+      this.requires.merge(conv.requires);
+      returnIDL = `
+        ${conv.body}
+        return returned;
+      `;
+    }
+
+    const treatNonObjectAsNull = utils.getExtAttr(this.idl.extAttrs, "TreatNonObjectAsNull");
+    const assertCallable = treatNonObjectAsNull ? "" : `
+      if (typeof value !== "function" && !treatNonObjectAsNull) {
```
ExE-Boss:
```suggestion
      if (typeof value !== "function") {
```
The&nbsp;`treatNonObjectAsNull`&nbsp;variable is&nbsp;not&nbsp;defined in&nbsp;the&nbsp;generated&nbsp;file, so&nbsp;it&nbsp;would&nbsp;cause a&nbsp;`ReferenceError` to&nbsp;be&nbsp;thrown.

lib/constructs/callback-function.js
```diff
@@ -0,0 +1,102 @@
...
+    if (treatNonObjectAsNull) {
+      this.str += "}";
+    }
+    this.str += `
+            ${returnIDL}
+          }
+          wrapper.construct = function(...args) {
+            ${argsToES}
+            let returned = Reflect.construct(value, args);
+            ${returnIDL}
```
ExE-Boss: This&nbsp;should&nbsp;probably also&nbsp;be&nbsp;wrapped in&nbsp;<code>if&nbsp;(typeof&nbsp;value&nbsp;===&nbsp;"function")</code> when&nbsp;<code>treatNonObjectAsNull&nbsp;===&nbsp;true</code>.

ExE-Boss: Actually, I just had a read through [the relevant spec](https://heycam.github.io/webidl/#construct-a-callback-function), and `[TreatNonObjectAsNull]` isn’t specified for constructor callback functions.

test/__snapshots__/test.js.snap
ExE-Boss: Always setting `args.length` to a constant is only valid for callbacks without optional arguments.

domenic:
This is semi-blocking the Streams Standard reference implementation from migrating to webidl2js. If anyone has time to finish this up let me know; otherwise I'll try to do it!

TimothyGu:
I won't have time to work on this in the near future, unfortunately. Please by all means work on this if it's blocking other things.

PR #125, opened 2019-05-27
Escape certain reserved identifiers

### LINKED ISSUE #117: The syntax for WebIDL2JSValueAsUnsupported is invalid
```webidl
interface URLSearchParamsCollection {
  [WebIDL2JSValueAsUnsupported=null] getter URLSearchParams? namedItem(DOMString name);
};
```

Here, `null` is tokenized as a terminal symbol `null` rather than an identifier [per the grammar](https://heycam.github.io/webidl/#idl-grammar):

>If the longest possible match could match one of the above named terminal symbols or one of the other terminal symbols from the grammar, it must be tokenized as the latter.

And thus cannot be parsed as:

```
ExtendedAttributeIdent ::
    identifier = identifier
```

Currently webidl2.js 13+ fails to parse it, so this should be fixed before upgrading to the latest version.

TimothyGu:
Fixes #117.

### DIFF
--- a/README.md
+++ b/README.md
@@ -395,7 +395,7 @@ It is currently used by [jsdom](https://github.com/tmpvar/jsdom) for classes whi
 
 ### `[WebIDL2JSValueAsUnsupported=value]`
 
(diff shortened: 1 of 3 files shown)


TimothyGu:
@domenic precisely.

PR #126, opened 2019-06-01
Add IDLHarness for interface tests

Sebmaster:
This finally gets us some behaviour-based tests in this repo instead of relying on manually checked snapshots.


test/harness.test.js
```diff
@@ -0,0 +1,170 @@
...
+    return vm.runInContext(
+      `(() => {
+        let exports = {};
+        const module = { exports };
+        (() => {${src}})();
+        return module.exports;
+      })();`,
+      sandbox,
+      { filename: resolved, lineOffset: 5, columnOffset: 16 }
+    );
```
ExE-Boss: It might be better to use [`vm.compileFunction(…)`](https://nodejs.org/api/vm.html#vm_vm_compilefunction_code_params_options) here.

PR #130, opened 2019-09-25
Add [CEReactions] and [HTMLConstructor] processors

pmdartus:
## Motivation

In order to implement [Custom Element in jsdom](https://github.com/jsdom/jsdom/pull/2548), we need to add support for `[CEReactions]` and `[HTMLConstructor]` IDL extended attributes. Since those 2 extended attributes require some knowledge about jsdom implementation details, we need a way for JSDOM to hook itself into the code generation.

## Changes

This PR introduces new 2 optional processor `processCEReactions` and `processHTMLConstructor`. Those processors receive 2 arguments: the parsed IDL property by webidl2 and the code generated by webidl2js, and expect to return some post-processed code.

Here is an example of how those new hooks would integrate into jsdom: [convert.js](https://github.com/pmdartus/jsdom/blob/pmdartus/custom-element/scripts/webidl/convert.js#L14-L49)

----

Note that this PR is already out of date because of the introduction of the new WebIDL `constructor` syntax. I opened another issue for the `constructor` syntax upgrade (https://github.com/jsdom/webidl2js/issues/131).


README.md
```diff
@@ -357,6 +359,8 @@ webidl2js is implementing an ever-growing subset of the Web IDL specification. S
 - `[TreatNullAs]`
 - `[Unforgeable]`
 - `[Unscopable]`
+- `[CEReactions]`
+- `[HTMLConstructor]`
```
TimothyGu: I'd open a new section for these two as "Web IDL extensions defined in HTML."

lib/constructs/interface.js
```diff
@@ -1024,72 +1025,88 @@ class Interface {
...
+        definePropertyTrap += "}";
       }
     } else {
       needFallback = true;
     }
     if (needFallback) {
       // Spec says to set configurable to true, but doing so will make Proxy's trap throw and also fail WPT.
       // if (!this.isGlobal) {
-      //   this.str += `
+      //   defineProperty += `
```
TimothyGu:
```suggestion
      //   definePropertyTrap += `
```

lib/transformer.js
```diff
@@ -144,6 +146,7 @@ class Transformer {
               break;
             }
 
+
```
TimothyGu:
```suggestion
```

lib/utils.js
```diff
@@ -32,6 +32,10 @@ function isGlobal(idl) {
   return Boolean(getExtAttr(idl.extAttrs, "Global"));
 }
 
+function hasCeReactions(idl) {
```
TimothyGu:
```suggestion
function hasCEReactions(idl) {
```

test/cases/CEReactions.webidl
```diff
@@ -0,0 +1,8 @@
+interface CEReactions {
+    [CEReactions] attribute DOMString attribute;
+    [CEReactions] void method();
+    
+    getter DOMString (DOMString name);
+    [CEReactions] setter void (DOMString name, DOMString value);
+    [CEReactions] deleter void (DOMString name);
+};
```
TimothyGu: nit: missing training new line. Also in HTMLConstructor.webidl.

test/__snapshots__/test.js.snap
```diff
@@ -1,5 +1,320 @@
...
+
+const impl = utils.implSymbol;
+
+class CEReactions {
+  constructor() {
+    throw new TypeError(\\"Illegal constructor\\");
+  }
+
+  method() {
+    // CEReaction pre steps
```
TimothyGu: This seems to be a bit inconsistent: the brand check is outside the code passed to the user-defined hook for accessors. The [spec](https://html.spec.whatwg.org/multipage/custom-elements.html#cereactions) says

> Operations, attributes, setters, or deleters annotated with the `[CEReactions]` extended attribute must run the following steps **in place of the ones specified in their description**

which seems to me that the brand/type checking and other Web IDL conversion code should not be included as part of the hook.

E.g., for deleters:

```diff
       deleteProperty(target, P) {
-        // CEReaction pre steps
 
         if (typeof P === \\"symbol\\") {
           return Reflect.deleteProperty(target, P);
         }
 
         if (target[impl][utils.supportsPropertyName](P) && !(P in target)) {
+        // CEReaction pre steps
           target[impl][utils.namedDelete](P);
+        // CEReaction post steps
           return true;
         }
 
         return Reflect.deleteProperty(target, P);
 
-        // CEReaction post steps
       },
```

README.md
TimothyGu: Perhaps also say that these are "supported" in the sense that the integrator can use hooks to properly modify their behavior?

README.md
```diff
@@ -108,8 +108,10 @@ transformer.generate("wrappers").catch(err => {
 
 The main module's default export is a class which you can construct with a few options:
 
-- `implSuffix`: a suffix used, if any, to find files within the source directory based on the IDL file name.
-- `suppressErrors`: set to true to suppress errors during generation.
+- `implSuffix`: a suffix used, if any, to find files within the source directory based on the IDL file name. (default: `""`)
+- `suppressErrors`: set to true to suppress errors during generation. (default: `false`)
+- `processCEReactions`: an optional hook to post-process the generated code for IDL operation and attributes annotated with the [`[CEReaction]`](https://html.spec.whatwg.org/multipage/custom-elements.html#cereactions) extended attributes. The processor expects the following signature `(idl: IdlOperation | IdlAttribute, code: string): string` where `idl` is the parsed [operation member](https://github.com/w3c/webidl2.js#operation-member) and [attribute member](https://github.com/w3c/webidl2.js#attribute-member) and `code` the generated accessor code.
```
TimothyGu:
```suggestion
- `processCEReactions`: an optional hook to post-process the generated code for IDL operation and attributes annotated with the [`[CEReaction]`](https://html.spec.whatwg.org/multipage/custom-elements.html#cereactions) extended attribute. The processor expects the following signature `(idl: IdlOperation | IdlAttribute, code: string): string` where `idl` is the parsed [operation member](https://github.com/w3c/webidl2.js#operation-member) and [attribute member](https://github.com/w3c/webidl2.js#attribute-member) and `code` the generated accessor code.
```

I'd also maybe say a few words on a) the default behavior (pass-through) as well as b) what the `code` parameter actually contains. An example here would be more than sufficient.

Same for `[HTMLConstructor]`.

test/__snapshots__/test.js.snap
domenic: Good catch. This is not as clear as it could be in the spec; see https://github.com/whatwg/html/issues/3879. But @TimothyGu is right on the intent of the spec.

domenic:
Thanks so much for the review @TimothyGu. I don't have much to add, just that overall I'm really happy with this approach and excited about what it represents (i.e., progress on custom elements in jsdom).

pmdartus:
@TimothyGu Ready for another pass of review.

domenic:
@pmdartus I pushed some doc tweaks; let me know what you think. I would also like to make the HTMLConstructor example slightly more realistic (like I did with the try/finally for CEReactions); if you have any ideas there, or samples of the jsdom code you're planning on uploading, let me know.

Do you think we should do #131 before or after landing this PR? My instinct is to try to finish a first draft of custom element support ASAP, so not wait on #131. But if you want to do that then I'll do my best to be speedy about reviews, now that my vacation is over :).

pmdartus:
Thanks for fixing my broken fr-english, the documentation looks way better now!

I think you are right, I would prefer to release a first draft of CE and get some initial feedback before tackling #131. In terms of ordering, I was thinking about working on #131 once #138 is merged.

domenic:
Oh, I see that https://github.com/jsdom/jsdom/pull/2548 is updated to use this already. I'll tweak the documentation to reflect how things work for [HTMLConstructor]. And then I'll try to work on reviewing/pushing tweaks for that!

I think it'll be best to merge this when the jsdom PR is approved, just in case we want to make any extra changes here as part of the review process over there.

test/__snapshots__/test.js.snap
```diff
@@ -862,6 +1181,103 @@ const Impl = require(\\"../implementations/Global.js\\");
 "
 `;
 
+exports[`HTMLConstructor.webidl 1`] = `
+"\\"use strict\\";
+
+const conversions = require(\\"webidl-conversions\\");
+const utils = require(\\"./utils.js\\");
+
+// HTMLConstructor
```
TimothyGu: According to how https://heycam.github.io/webidl/#overridden-constructor-steps is used (as part of the definition for `[HTMLConstructor]`), it seems like it would be more appropriate to pass the body of the constructor to the hook rather than the entire interface. However happy to have this if it turns out to be necessary for the implementation of CE in jsdom.

pmdartus: Agreed, passing the constructor would be ideal. 

In the case `[HTMLConstructor]` a new constructor is needed per realm to capture the global object. In order to get around this, I inject an install method to the exported `iface` object. This is why the entire module is passed and not only the constructor.

You can see how this hook is used on the CE PR: https://github.com/jsdom/jsdom/pull/2548/files#diff-4c420ef958793615b9bf0a6fd47a1fe1R29-R50

test/__snapshots__/test.js.snap
```diff
@@ -1,5 +1,324 @@
...
+          return Reflect.deleteProperty(target, P);
+        }
+
+        if (target[impl][utils.supportsPropertyName](P) && !(P in target)) {
+          // CEReactions pre steps
+
+          target[impl][utils.namedDelete](P);
+          return true;
+
+          // CEReactions post steps
```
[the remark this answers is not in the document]
ExE-Boss: That’s why this should probably be wrapped in a `try‑finally` block.

test/test.js
```diff
@@ -12,7 +12,21 @@ const implsDir = path.resolve(__dirname, "implementations");
 const outputDir = path.resolve(__dirname, "output");
 
 beforeAll(() => {
-  const transformer = new Transformer();
+  const transformer = new Transformer({
+    processCEReactions(_, code) {
+      return `
+        // CEReactions pre steps
+        ${code}
```
ExE-Boss:
```suggestion
        try {
          ${code}
        } finally {
```

test/test.js
```diff
@@ -12,7 +12,21 @@ const implsDir = path.resolve(__dirname, "implementations");
...
+        // CEReactions post steps
```
ExE-Boss:
```suggestion
          // CEReactions post steps
        }
```

test/__snapshots__/test.js.snap
TimothyGu: Great. Thanks for explaining.

TimothyGu:
@pmdartus Looks like this now requires some additional rebasing. Feel free to merge otherwise.

pmdartus:
@domenic @TimothyGu I made some last-minute changes to the processor interface and updated the documentation. It would be great if you can give a look at it one last time before I merge it.

README.md
```diff
@@ -121,6 +122,57 @@ The transformer will also generate a file named `utils.js` inside the wrapper cl
...
+
+Both hooks have the signature `(code) => replacementCode`, where:
+
+- `code` is the code generated by webidl2js normally, for calling into the impl class.
+
+- `replacementCode` is the new code that will be output in place of `code` in the wrapper class.
+
+If either hook is omitted, then the code will not be replaced, i.e. the default is equivalent to `(code) => code`.
+
+Both hooks also some utility methods that are are accessible via `this`:
```
TimothyGu: Maybe also mention the variables that are in scope for the replacement code? e.g., `globalObject` being the object passed to `install()`. (Though I don’t find any other variables being consistent throughout the different contexts.)

TimothyGu: Missing a “have” after “also”? Duplicate “are”?

README.md
```diff
@@ -121,6 +122,57 @@ The transformer will also generate a file named `utils.js` inside the wrapper cl
...
+      // CEReactions pre steps
+      try {
+        ${code}
+      } finally {
+        // CEReactions post steps
+      }
+    `;
+  },
+  processHTMLConstructor(code) {
+    const identifier = this.addImport("../HTMLConstructor", "HTMLConstructor");
```
TimothyGu: It’d be nice to show how `identifier` could be used in the returned code.

test/test.js
```diff
@@ -12,7 +12,28 @@ const implsDir = path.resolve(__dirname, "implementations");
 const outputDir = path.resolve(__dirname, "output");
 
 beforeAll(() => {
-  const transformer = new Transformer();
+  const transformer = new Transformer({
+    processCEReactions(code) {
+      const preSteps = this.addImport("../CEReactions", "preSteps");
+      const postSteps = this.addImport("../CEReactions", "postSteps");
```
ExE-Boss: Maybe consider:
```suggestion
      const csReactions = this.addImport("../CEReactions");
```

Since&nbsp;imports aren't&nbsp;de‑duplicated.

test/test.js
```diff
@@ -12,7 +12,28 @@ const implsDir = path.resolve(__dirname, "implementations");
...
+
+      return `
+        ${preSteps}(globalObject);
```
ExE-Boss:
```suggestion
        ${ceReactions}.preSteps(globalObject);
```

test/test.js
```diff
@@ -12,7 +12,28 @@ const implsDir = path.resolve(__dirname, "implementations");
...
+        try {
+          ${code}
+        } finally {
+          ${postSteps}(globalObject);
```
ExE-Boss:
```suggestion
          ${ceReactions}.postSteps(globalObject);
```

README.md
```diff
@@ -121,6 +122,66 @@ The transformer will also generate a file named `utils.js` inside the wrapper cl
...
+
+```js
+"use strict";
+const WebIDL2JS = require("webidl2js");
+
+const transformer = new WebIDL2JS({
+  implSuffix: "-impl",
+  processCEReactions(code) {
+    const ceReactionsPreSteps = this.addImport("../ce-reactions", "ceReactionsPreSteps");
+    const ceReactionsPostSteps = this.addImport("../ce-reactions", "ceReactionsPostSteps");
```
ExE-Boss:
```suggestion
    const ceReactions = this.addImport("../ce-reactions");
```

README.md
```diff
@@ -121,6 +122,66 @@ The transformer will also generate a file named `utils.js` inside the wrapper cl
...
+
+    return `
+      ${ceReactionsPreSteps}(globalObject);
```
ExE-Boss:
```suggestion
      ${ceReactions}.preSteps(globalObject);
```

README.md
```diff
@@ -121,6 +122,66 @@ The transformer will also generate a file named `utils.js` inside the wrapper cl
...
+      try {
+        ${code}
+      } finally {
+        ${ceReactionsPostSteps}(globalObject);
```
ExE-Boss:
```suggestion
        ${ceReactions}.postSteps(globalObject);
```

test/test.js
TimothyGu: Sounds reasonable. Fixed.

TimothyGu:
Rebased and added a bit of documentation change as well as @ExE-Boss's comments. @domenic I'll let you do the honors?

PR #132, opened 2019-09-26
Retire Object.getOwnPropertyDescriptors polyfill

### LINKED ISSUE #112: Raise minimum support level to Node 8
Then we can get rid of the getOwnPropertyDescriptors polyfill

pmdartus:
## Changes

* Remove `Object.getOwnPropertyDescriptors` polyfill
* Raise min Node version to 8

Fixes #112.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,5 @@
 language: node_js
 node_js:
-  - "6"
   - "8"
   - stable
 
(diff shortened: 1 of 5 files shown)


TimothyGu:
This will have to accompany a corresponding jsdom update.

domenic:
jsdom already requires v8, so this won't be breaking there. @pmdartus, can you remove Node v6 from the Travis config too?

PR #135, opened 2019-10-12
Remove old-style mixins

### LINKED ISSUE #111: Remove old-style mixins (implements)
Now that we have actual mixins, we should remove all that code.

pmdartus:
Fixes #111.


ExE-Boss:
There’s still some dead code left from this, which I’m cleaning up in <https://github.com/jsdom/webidl2js/pull/171>.

PR #138, opened 2019-10-13
WIP upgrade webidl2

domenic:
Fixes so far:

* Removed `[NoInterfaceObject]` support and made a single `[Exposed]` mandatory to align with ongoing/upcoming Web IDL spec changes. Fixed how `[Exposed]` arguments are processed as that changed in webidl2.
* Stopped checking for `.arguments` when checking `[Global]` validity; I'm not sure what that check was for and it doesn't seem to work anymore anyway.
* Updated to check the `.special` property instead of the `.getter`, `.setter`, `.deleter`, `.stringifier`, `.static` properties
* Changed to check for falsy `name` properties for indexed/named getters/setters, instead of `null`. webidl2 is now sometimes returning undefined and sometimes returning the empty string (I think via the prototype chain? There is no `name` own property.)

Tests still failing:

* Escaped (underscore-prefixed) name handling got changed in webidl2, causing various failures due to `_analyzeMembers (lib/constructs/interface.js:367:15)` which seems to prohibit any underscores from appearing. I'm unsure what the intent of the original code was, or the intent of the webidl2 changes, which seem likely to be those in https://github.com/w3c/webidl2.js/pull/229. This is causing lots of early failures which might mask other issues.
* `sequence<sequence<USVString>>` in `URLSearchParams.webidl` is being treated as something like `sequence<USVString>`. I fixed several similar failures, which were caused by webidl2 switching `idlType` from the inner type to an array containing the inner type. I'm unsure why this one remains.

Help very much appreciated from anyone who is able to figure these remaining things out.

/cc @saschanaz

### DIFF
--- a/README.md
+++ b/README.md
@@ -175,7 +175,7 @@ This export is the wrapper class interface, suitable for example for putting on 
 
 #### `expose`
 
-This export contains information about where an interface is supposed to be exposed as a property. It takes into account the Web IDL extended attribute `[Expose]` to generate a data structure of the form:
+This export contains information about where an interface is supposed to be exposed as a property. It takes into account the Web IDL extended attribute `[Exposed]` to generate a data structure of the form:
 
 ```js
 {
--- a/README.md
+++ b/README.md
@@ -175,7 +175,7 @@ This export is the wrapper class interface, suitable for example for putting on 
 
 #### `expose`
 
-This export contains information about where an interface is supposed to be exposed as a property. It takes into account the Web IDL extended attributes `[Expose]` and `[NoInterfaceObject]` to generate a data structure of the form:
+This export contains information about where an interface is supposed to be exposed as a property. It takes into account the Web IDL extended attribute `[Expose]` to generate a data structure of the form:
 
 ```js
 {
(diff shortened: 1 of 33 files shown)


saschanaz:
>     * Escaped (underscore-prefixed) name handling got changed in webidl2, causing various failures due to `_analyzeMembers (lib/constructs/interface.js:367:15)`

I think that code is not useful anymore as https://github.com/w3c/webidl2.js/pull/403 made it throw early.

domenic:
Awesome, thanks! With that hint I got a decent bit more working. What's left is still underscore-related; it seems like previously underscores were stripped and now they are retained? So jsdom is putting underscores in the output for `_const` and `_null`.

saschanaz:
That looks like a bug, I will investigate.

domenic:
@saschanaz still one failure, where

```webidl
[Exposed=Window]
interface _UnderscoredProperties {
  const byte _const = 42;
  attribute byte _attribute;
  static void _static(DOMString _void);
  void _operation(sequence<DOMString> _sequence);
};
```

seems to generate `_const`, instead of `const`, for the `name` of the constant member.

PR #140, opened 2019-10-20
Constructor & prototype reform

### LINKED ISSUE #133: Constructor & prototype reform
## Context 

Currently, the constructor and prototype chains are shared between all the realms. This approach has multiple downstream impacts:
* on the maintainer side, it requires hacks to implements interfaces that need a reference to the current global object, eg: [`[HTMLConstructor]`](https://html.spec.whatwg.org/#htmlconstructor). WebIDL2JS also had to implement the [`[WebIDL2JSFactory]`](https://github.com/jsdom/webidl2js#webidl2jsfactory) IDL extended attribute to generate wrapper class per window object.
* on the consumer side, it is unexpected for developers that all constructors and prototypes are shared, especially when they use a test framework abstracting jsdom. I personally spent a lot of time figuring out the best way to apply polyfill in the context of jsdom.

While recreating a brand new wrapper class for each new realm won't scale, I think there might be other alternatives worth exploring.

## Approach 1: New constructor / Same prototype

This approach consists of recreating a brand new constructor per realm and reuse the same prototype. WebIDL2JS would generate for all the interfaces annotated with the `[Exposed]` IDL extended attribute would an `install` method that would be in charge of creating the interface in the context of the realm.

```js
class TestGlobal extends BaseGlobal { /* ... */ }

const iface = {
    /* ... */

    install(globalObject, privateData) {
        const Test = function() {
            return iface.setup(Object.create(TestGlobal.prototype), [], privateData);
        };
        // Apply static properties if necessary

        Object.setPrototypeOf(Test, TestGlobal.prototype);

        Object.defineProperty(globalObject, "Test", {
            configurable: true,
            overridable: true,
            value: Test
        });

        // Apply [NamedConstructor] if necessary
    }
}
module.export = iface;
```

You can find here an experimental branch with the code change: [pmdartus/constructor](https://github.com/jsdom/webidl2js/compare/master...pmdartus:pmdartus/constructor)

**Pros:**
- Really light-weight approach.
- No need for `[WebIDL2JSFactory]` anymore.
- The new install method can be used to implement `[NamedConstructor]`.
- Minimal change needed on jsdom

**Cons:**
- It doesn't solve the polyfill issue.
- With this approach the only the `(new Test()).constructor` points to `TestGlobal` instead of `Test`, since the prototype is shared. When running the experimental WebIDL2JS against jsdom test, this issue made 5 tests failed.

## Approach 2: New constructor / New prototype

In this approach, a brand new constructor and prototype chain is created per realm. Instead of creating a brand new wrapper class per realm, like `[WebIDL2JSFactory]`, we would instead only create a new prototype chain but reuse the existing property descriptors. It also reuses the concept of `install` showed in proposal 1.

The assumption here is that the performance cost of the creation of a prototype chain while reusing the existing descriptors is way lower than re-creating a brand new class per realm.

```js
class TestGlobal extends BaseGlobal { /* ... */ }

const iface = {
    /* ... */

    install(globalObject, privateData) {
        const Test = function() {
            return iface.setup(Object.create(TestGlobal.prototype), [], privateData);
        };
        // Apply static properties if necessary

        const proto = Object.create(globalObject.Base.prototype);
        Object.defineProperties(proto, Object.getOwnProperties(TestGlobal.prototype));

        Object.setPrototypeOf(Test, proto);

        Object.defineProperty(globalObject, "Test", {
            configurable: true,
            overridable: true,
            value: Test
        });

        // Apply [NamedConstructor] if necessary
    }
}
module.export = iface;
```

**Pros:**
- No need for `[WebIDL2JSFactory]` anymore.
- The new install method can be used to implement `[NamedConstructor]`.
- Solves the polyfill issue.
- Fully spec compliant.

**Cons:**
- Heavier solution. Would need to do a POC to evaluate the actual performance penalty of this approach.
- Requires changes to jsdom. Now that we have a brand new prototype per realm, we need to pass the global object for every `create()` and `createImpl()` invocations, for it to retrieve the right prototype associated with the realm.

----

Before diving deeper into this, I would like getting some initial feedback on both approaches and see if it worth pursuing.

/cc @domenic @TimothyGu @Sebmaster @Zirro

pmdartus:
This PR adds the capability for webidl2js to generate a brand new constructor and prototype chain per realm. Fixes #133

> Note: The code change below doesn't reflect yet the proposed API changes. I want to make sure that we agree on the APIs before making all the changes.

### Interface wrapper `install` method introduction

The `install` method takes care of creating a new constructor and a new prototype chain for a realm. This method is only generated on WebIDL interfaces. This method takes a single parameter, `globalObject` which is the global object attached to the realm.

### Interface wrapper `create` and `createImpl` signature change

Now that we have a brand new constructor and prototype per realm, we need to create wrapper instances with the right constructor. In order to do so, I am proposing passing the `globalObject` as the first parameter to `create(globalObject, constructorArgs, privateData)` and `createImpl(globalObject, constructorArgs, privateData)`.

### Interface implementation `constructor` signature change

In the proof-of-concept, I originally passed the `globalObject` to the implementation via the `privateData` to minimize the amount of code change in jsdom. Now thinking about it, a preferable approach making the `constructor` signature mirror the new `create` and `createImpl` signature: `constructor(globalObject, constructorArgs, privateData)`.

### Interface wrapper `interface` and `expose` property removal

I don't make sense to expose the shared interface wrapper class anymore.

### Custom WebIDL extended attribute `[WebIDL2JSFactory]` removal

Not needed anymore with this proposal.

## Open questions

- I am wondering if it is a good time to add support for the `[Exposed]` WebIDL ext attribute since JSDOM doesn't need it right? The proposed `install` method would blindly attach the constructor to the passed global object.


domenic:
This all sounds reasonable to me.

Currently `[Exposed]` is "supported" in that the metadata is added to the output, which theoretically a consumer like jsdom could use (especially if we added Worker support). I am OK removing this support for now (with the appropriate readme updates).

Eventually we may want to move in the direction of #101 where you can do something like

```js
require('./generated/FooInterface.js').expose("Window", windowObj);
```

or even

```js
require('./generated/bundle-entry.js').exposeAllInterfaces("Window", windowObj);
```

But that is a separate project. So in the meantime let's just remove the expose metadata entirely and have `install()` (or `expose()`? not sure which name I like better---up to you) put it on the passed object unconditionally.

lib/constructs/interface.js
```diff
@@ -1120,14 +1120,22 @@ class Interface {
     }
 
     this.str += `
-      create(constructorArgs, privateData) {
-        let obj = Object.create(${this.name}.prototype);
-        obj = this.setup(obj, constructorArgs, privateData);
+      create(globalObject, constructorArgs, privateData) {
+        if (globalObject[ctorRegistry] === undefined) {
+          throw new TypeError('Invalid global object');
```
domenic:
```suggestion
          throw new Error('Internal error: invalid global object');
```

lib/constructs/interface.js
```diff
@@ -1120,14 +1120,22 @@ class Interface {
...
+        }
+
+        const ctor = globalObject[ctorRegistry]["${this.name}"];
+        if (ctor === undefined) {
+          throw new TypeError('Constructor ${this.name} is not installed on the passed global object');
```
domenic:
```suggestion
          throw new Error('Internal error: constructor ${this.name} is not exposed on the passed global object');
```

lib/constructs/interface.js
domenic: I think these errors should never come up unless we mess up as jsdom maintainers. Right? So let's prefix them with "Internal error".

test/__snapshots__/test.js.snap
```diff
@@ -201,6 +210,38 @@ const iface = {
...
+
+  install(globalObject) {
+    const DOMImplementation = function() {
+      throw new TypeError(\\"Illegal constructor\\");
+    };
+
+    const proto = {};
+
+    Object.defineProperties(proto, {
+      ...Object.getOwnPropertyDescriptors(iface.interface.prototype),
```
domenic: Interesting. I'm unsure whether this will actually be faster than just doing `globalObject.DOMImplementation = { ... a bunch of generated code ... }`. And it seems like it'll have issues like `window.DOMImplementation.prototype.createDocument instanceof window.Function === false`.

But, it seems reasonable for now though as a smaller delta from what we're doing currently...

ExE-Boss:
Since this will introduce breaking changes, does it make sense to address #59 as well?

test/__snapshots__/test.js.snap
pmdartus: As far as I know, creating new function every `install` invocation would have worse runtime performance characteristics than creating a brand new class constructor per `install` (like with the `[WebIDL2JSFactory]`).

I didn't want to make too many changes into a single PR, but my next goal is to get rid of the `class` and generate a property descriptor for the prototype and a property descriptor for the constructor. With this PR, the class is only used to retrieve its property descriptors. The idea would be to generate something like this in the future:

```js
const constructorDescriptors = {
    /* generated descriptors */
};
const prototypeDescriptors = {
    /* generated descriptors */
};

function install() {
    const ctor = function Node() {
        throw new TypeError('Illegal constructor');
    };
    Object.defineProperties(ctor, constructorDescriptors);

    const proto = {};
    Object.defineProperties(proto, constructorDescriptors);

    Object.defineProperty(ctor, 'prototype', { value: proto });
    Object.defineProperty(proto, 'constructor', { value: ctor });
}
```

pmdartus:
I am finally done updating jsdom to consume the new APIs 🎉. You can see the new APIs in actions here: https://github.com/jsdom/jsdom/pull/2691

test/__snapshots__/test.js.snap
domenic: I'd like to not pursue any complicated attempts at sharing descriptors (and, notably, sharing the functions that are the `value` property of descriptors) without some benchmarks showing that's better than just creating a new copy each time. Engines are smart; I suspect that all these descriptor tricks could actually be less optimized than just

```js
function install() {
  class Node {
    // everything goes inside here
  }
}
```

lib/constructs/interface.js
```diff
@@ -1436,18 +1404,98 @@ class Interface {
...
+    ];
+    if (staticPropertyNames.length) {
       this.str += `
-        module.exports = {
-          createInterface: function (defaultPrivateData = {}) {
+        Object.defineProperties(ctor, {
+      `;
+
+      for (const staticProperty of staticPropertyNames) {
+        // TODO: Improve this code. It's a really bad hack.
```
domenic: What's the plan here?

test/__snapshots__/test.js.snap
```diff
@@ -170,25 +167,31 @@ const iface = {
...
+      throw new Error(\\"Internal error: invalid global object\\");
+    }
+
+    const ctor = globalObject[ctorRegistry][\\"DOMImplementation\\"];
+    if (ctor === undefined) {
+      throw new Error(\\"Internal error: constructor DOMImplementation is not installed on the passed global object\\");
+    }
+
+    let obj = Object.create(ctor.prototype);
+    obj = this.setup(obj, globalObject, constructorArgs, privateData);
```
domenic: Why `this.` here and `iface.` below? Maybe `iface.` in both would be better.

test/__snapshots__/test.js.snap
```diff
@@ -198,9 +201,37 @@ const iface = {
...
+        value: ctor
+      }
+    });
+    Object.defineProperty(ctor, \\"prototype\\", {
+      writable: false,
+      value: proto
+    });
+
+    if (globalObject[ctorRegistry] === undefined) {
+      globalObject[ctorRegistry] = {};
```
domenic: Maybe `Object.create(null)` would be better, to avoid treating e.g. `hasOwnProperty` as a constructor.

lib/constructs/interface.js
pmdartus: Fixed.

pmdartus:
I came to the realization that there is more nuance to the `interface` property removal from the generated wrapper. Packages like jsdom/whatwg-url rely on the [interface property exposure](https://github.com/jsdom/whatwg-url/blob/master/lib/public-api.js#L3-L4).

The following changes are needed to accommodate the new API and keep backward compatibility:
- install the generated wrapper on a shared global object: [code](https://github.com/jsdom/whatwg-url/compare/master...pmdartus:pmdartus/constructor-reform-pt2#diff-c108551d69a9d494e93b86a6c044b69fR9-R11)
- expose the generated wrapper out of the package for jsdom to install a new prototype per global object: [code](https://github.com/jsdom/whatwg-url/compare/master...pmdartus:pmdartus/constructor-reform-pt2#diff-c108551d69a9d494e93b86a6c044b69fR6-R7)

After thinking more about it, making `install` invocation mandatory to get the wrapper class vs. exposing it via `interface` is the best approach. Even if the API is way less friendly, this approach prevents prototype pollution.

I wanted to raise this concern before moving forward with this PR. I would be happy to make the required changes if you think that we should bring back the `interface` property.

pmdartus:
@domenic you were right the inline class declaration has the same performance characteristics or outperforms in some cases, the synthetic constructor creation. You can find the performance results below and the code [here](https://github.com/jsdom/webidl2js/compare/master...pmdartus:pmdartus/inline-constructor).

@TimothyGu Do you still have the code somewhere with your initial investigation (https://github.com/jsdom/webidl2js/issues/133#issuecomment-539402929)? I would like to better understand why your approach didn't work in the first place.

----

<details>
 <summary>Performance results</summary>

|                   **Test name**                   | **Synthetic class (ops/sec)** | **Inline constructor (ops/sec)** | **Delta** |
|:-------------------------------------------------:|:-----------------------------:|:--------------------------------:|:---------:|
| dom/compare-document-position/compare ancestor    | 19,712                        | 20,731                           | 5.17%     |
| dom/compare-document-position/compare descendant  | 19,632                        | 20,530                           | 4.57%     |
| dom/compare-document-position/compare siblings    | 1,116,531                     | 1,132,943                        | 1.47%     |
| dom/construction/createComment                    | 1,398,768                     | 1,243,397                        | -11.11%   |
| dom/construction/createDocumentFragment           | 1,385,996                     | 1,366,932                        | -1.38%    |
| dom/construction/createElement                    | 200,420                       | 203,415                          | 1.49%     |
| dom/construction/createEvent                      | 193,832                       | 197,936                          | 2.12%     |
| dom/construction/createNodeIterator               | 1,359,636                     | 1,314,331                        | -3.33%    |
| dom/construction/createProcessingInstruction      | 687,056                       | 674,050                          | -1.89%    |
| dom/construction/createTextNode                   | 1,328,920                     | 1,422,536                        | 7.04%     |
| dom/inner-html/tables                             | 0.3                           | 0.33                             | 10.00%    |
| dom/named-properties/setAttribute                 | 3,404                         | 2,828                            | -16.92%   |
| dom/tree-modification/appendChild: many parents   | 23.41                         | 23.33                            | -0.34%    |
| dom/tree-modification/appendChild: many siblings  | 238,392                       | 225,715                          | -5.32%    |
| dom/tree-modification/appendChild: no siblings    | 216,586                       | 213,095                          | -1.61%    |
| dom/tree-modification/insertBefore: many siblings | 207,314                       | 212,118                          | 2.32%     |
| dom/tree-modification/removeChild: many parents   | 121                           | 135                              | 11.57%    |
| dom/tree-modification/removeChild: many siblings  | 119,345                       | 135,362                          | 13.42%    |
| dom/tree-modification/removeChild: no siblings    | 76,650                        | 123,186                          | 60.71%    |
| html/parsing/text                                 | 13.07                         | 12.59                            | -3.67%    |
| jsdom/api/new JSDOM() defaults                    | 178                           | 231                              | 29.78%    |
| jsdom/api/new JSDOM() with many elements          | 25.33                         | 23.97                            | -5.37%    |

</details>

domenic:
> The following changes are needed to accommodate the new API and keep backward compatibility:

Thanks for outlining this. This is indeed more subtle. I think these changes are reasonable, although I'd like to name them something like `webidl2jsURL` instead of `URLWrapper`.

pmdartus:
Based on the performance investigation results, I decided to move forward with the inline class declaration. The PR is ready for another round of review.

lib/output/utils.js
```diff
@@ -12,6 +12,7 @@ function hasOwn(obj, prop) {
 const wrapperSymbol = Symbol("wrapper");
 const implSymbol = Symbol("impl");
 const sameObjectCaches = Symbol("SameObject caches");
+const ctorRegistrySymbol = Symbol.for("constructor registry");
```
ExE-Boss: This should probably be something more like:
```suggestion
const ctorRegistrySymbol = Symbol.for("[webidl2js] constructor registry");
```

Since [`Symbol.for(…)`](https://developer.mozilla.org/docs/Web/JavaScript/Reference/Global_Objects/Symbol/for) returns a&nbsp;shared symbol, it’s&nbsp;better to&nbsp;choose a&nbsp;name that is&nbsp;less likely to&nbsp;accidentally conflict.

domenic: Is there a particular reason to use Symbol.for() here instead of Symbol()?

pmdartus: I choosed to use the `Symbol.for()` to have a single constructor registry per global object. The interfaces may originate from different packages (jsdom, whatwg-url, ...), and if we were to use `Symbol()` it would create on the same global object one constructor registry per package.

ExE-Boss: Well, the&nbsp;symbol’s&nbsp;name should&nbsp;still be&nbsp;prefixed to&nbsp;avoid accidental&nbsp;conflicts with&nbsp;symbols from&nbsp;unrelated&nbsp;packages.

pmdartus: Done.

orneryd:
When will this be merged?

domenic:
When we have free time.

PR #141, opened 2019-10-22
refactor(interface): Use default export for implementation

### LINKED ISSUE #59: Can implementation classes default-export implementation, instead of having a single export?

ExE-Boss:
# ⚠️ **WARNING:** BREAKING CHANGE

Fixes #59

---

This changes the implementation class syntax to allow CommonJS and ES2015 style default exports:
```js
// CommonJS
/** @type {WeakMap<Foo, PrivateData>} */
const fooPrivateData = new WeakMap();

class Foo {
	// implementation
}
module.exports = exports = Foo;

export.init = function(instance, privateData) {
	fooPrivateData.set(instance, privateData);
}
```

```js
// ES2015
/** @type {WeakMap<Foo, PrivateData>} */
const fooPrivateData = new WeakMap();

export default class Foo {
	// implementation
}

export function init(instance, privateData) {
	fooPrivateData.set(instance, privateData);
}
```

---

## TODO:
- [x] Add tests for the `importStar(…)` helper.
- [x] Update documentation

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,20 +15,33 @@ interface SomeInterface {
 combined with the JavaScript implementation class file
 
 ```js
-exports.implementation = class SomeInterfaceImpl {
+module.exports = class SomeInterfaceImpl {
   add(x, y) {
     return x + y;
   }
 };
 ```
 
+> Note: It's also possible to use ES2015 module default export syntax:
+>
+> ```js
+> export default class SomeInterfaceImpl {
+>   add(x, y) {
+>     return x + y;
+>   }
+> };
+> ```
+
 will generate a JavaScript wrapper class file roughly like this:
 
 ```js
 const conversions = require("webidl-conversions");
-const impl = require("./utils.js").implSymbol;
+const utils = require("./utils.js");
+const impl = utils.implSymbol;
 
-const Impl = require("./SomeInterface-impl.js").implementation;
+// utils.importStar is roughly equivalent to Babel's _interopRequireWildcard()
+// and TypeScript's __importStar() functions.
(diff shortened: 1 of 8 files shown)


README.md
```diff
@@ -15,20 +15,33 @@ interface SomeInterface {
...
 ```js
-exports.implementation = class SomeInterfaceImpl {
+module.exports = class SomeInterfaceImpl {
   add(x, y) {
     return x + y;
   }
 };
 ```
 
+> Note: It's also possible to use ES2015 module default export syntax:
```
domenic: This is not true; jsdom only operates on CommonJS modules.

lib/output/utils.js
```diff
@@ -74,6 +74,52 @@ const namedSetNew = Symbol("named property set new");
 const namedSetExisting = Symbol("named property set existing");
 const namedDelete = Symbol("named property delete");
 
+/**
```
domenic: I am not interested in this addition. Please remove it.

ExE-Boss: Which? The&nbsp;JSDoc or&nbsp;the&nbsp;entire `utils.importStar(…)`&nbsp;function?

README.md
ExE-Boss: Well, you need to transpile them first.

lib/output/utils.js
domenic: The importStar function itself.

README.md
domenic: Yes. That is not a supported scenario, so please remove it.

PR #144, opened 2019-11-22
test(utils): Add tests for utils

ExE-Boss:
This&nbsp;adds&nbsp;tests for&nbsp;the&nbsp;`utils.js`&nbsp;file.

### DIFF
--- a/test/utils.test.js
+++ b/test/utils.test.js
@@ -1,52 +1,29 @@
-// eslint-disable-next-line spaced-comment
-/* eslint-env node, jest */
-/* global BigInt */
-
 "use strict";
 
 const utils = require("../lib/output/utils");
 
 describe("utils.js", () => {
   describe("isObject", () => {
-    const BIG_INT_PLACEHOLDER = "BigInt placeholder";
-
-    /** @type {Array<string | number | bigint | boolean | symbol | null | undefined>} */
     const primitives = [
       123,
       "string",
       Symbol.iterator,
       true,
-      BIG_INT_PLACEHOLDER,
       null,
       undefined
     ];
 
-    // eslint-disable-next-line new-cap, valid-typeof
-    if (typeof BigInt === "function" && typeof BigInt("123") === "bigint") {
-      primitives.splice(
-        primitives.indexOf(BIG_INT_PLACEHOLDER),
-        1,
-        // eslint-disable-next-line new-cap
-        BigInt("123")
-      );
-    }
-
     for (const primitive of primitives) {
-      /** @type {"string" | "number" | "bigint" | "boolean" | "symbol" | "null" | "undefined"} */
-      let type = primitive === null ? "null" : typeof primitive;
-
-      let testPrimitive = test;
-
-      if (primitive === BIG_INT_PLACEHOLDER) {
-        // BigInt doesn't exist on Node 8, so we have to assign this manually:
-        type = "bigint";
-        testPrimitive = test.skip;
-      }
-
-      testPrimitive(type, () => {
+      test(primitive === null ? "null" : typeof primitive, () => {
         expect(utils.isObject(primitive)).toBe(false);
       });
     }
+
+    const testBigInt = typeof BigInt === "function" && typeof BigInt("123") === "bigint" ? test : test.skip;
+
+    testBigInt("bigint", () => {
+      expect(utils.isObject(BigInt(123))).toBe(false);
+    });
 
     test("object", () => {
       expect(utils.isObject({})).toBe(true);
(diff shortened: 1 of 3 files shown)


test/utils.test.js
```diff
@@ -0,0 +1,59 @@
...
+      Symbol.iterator,
+      true,
+      BIG_INT_PLACEHOLDER,
+      null,
+      undefined
+    ];
+
+    // eslint-disable-next-line new-cap, valid-typeof
+    if (typeof BigInt === "function" && typeof BigInt("123") === "bigint") {
+      primitives.splice(
```
pmdartus: Instead of creating adding a placeholder and replacing the placeholder if `BigInt` is defined, what do you think about just pushing a `BigInt` at the end of the `primitives` array if it is defined? I think it would greatly simplify the test code.

```js
const primitives = [ /*...*/ ];

if (typeof BigInt === "function") { 
  primitives.push(BigInt(123));
}
```

ExE-Boss: Well, I&nbsp;want the&nbsp;test&nbsp;logs to&nbsp;show&nbsp;that the&nbsp;`BigInt`&nbsp;test was&nbsp;skipped.

ExE-Boss:
Can&nbsp;this be&nbsp;merged?

test/utils.test.js
```diff
@@ -0,0 +1,59 @@
+// eslint-disable-next-line spaced-comment
+/* eslint-env node, jest */
+/* global BigInt */
```
TimothyGu: Is there something we could do to avoid having to do this? Upgrading ESLint / changing ESLint configs?

test/utils.test.js
```diff
@@ -0,0 +1,59 @@
...
+
+"use strict";
+
+const utils = require("../lib/output/utils");
+
+describe("utils.js", () => {
+  describe("isObject", () => {
+    const BIG_INT_PLACEHOLDER = "BigInt placeholder";
+
+    /** @type {Array<string | number | bigint | boolean | symbol | null | undefined>} */
```
TimothyGu: Please remove JSDoc comments.

test/utils.test.js
TimothyGu: In that case, let's just do it the old fashioned way. Remove the `BIG_INT_PLACEHOLDER` trickery and instead just do

```js
const primitives = […];
for (const primitive of primitives) { … }

if (typeof BigInt === "function") { 
  test("bigint", () => {
    expect(utils.isObject(BigInt(123))).toBe(false);
  });
} else {
  test.skip("bigint");
}
```

test/utils.test.js
```diff
@@ -0,0 +1,59 @@
+// eslint-disable-next-line spaced-comment
```
TimothyGu: Is this needed given our ESLint configuration?

test/utils.test.js
```diff
@@ -0,0 +1,59 @@
...
+      123,
+      "string",
+      Symbol.iterator,
+      true,
+      BIG_INT_PLACEHOLDER,
+      null,
+      undefined
+    ];
+
+    // eslint-disable-next-line new-cap, valid-typeof
```
TimothyGu: Please add `BigInt` to https://github.com/jsdom/webidl2js/blob/701f958b609c9571c95c47998fdaf4c2aee97a28/.eslintrc.json#L165 rather than using `new-cap`.

test/utils.test.js
```diff
@@ -0,0 +1,59 @@
...
+    if (typeof BigInt === "function" && typeof BigInt("123") === "bigint") {
+      primitives.splice(
+        primitives.indexOf(BIG_INT_PLACEHOLDER),
+        1,
+        // eslint-disable-next-line new-cap
```
TimothyGu: Ditto.

test/utils.test.js
ExE-Boss: Actually, this&nbsp;is&nbsp;unnecessary as&nbsp;of&nbsp;**ESLint&nbsp;v6.2.0**.

Turns&nbsp;out that&nbsp;I&nbsp;had an&nbsp;outdated&nbsp;**ESLint** version&nbsp;installed when&nbsp;I&nbsp;originally wrote&nbsp;this.

test/utils.test.js
ExE-Boss: Actually, `test.skip`&nbsp;also&nbsp;takes the&nbsp;test&nbsp;case function&nbsp;argument.

TimothyGu:
Do you plan to add tests for any other utility functions?

PR #146, opened 2019-12-07
Avoid using instanceof for BufferSource types

domenic:
Given jsdom's move to separate the Node.js globals from the JSDOM globals, instanceof is no longer reliable in this fashion.

Helps with https://github.com/jsdom/jsdom/pull/2743.

---

Probably I should also fix webidl-conversions before pulling this.

### DIFF
--- a/lib/types.js
+++ b/lib/types.js
@@ -195,13 +195,13 @@ function generateTypeConversion(ctx, name, idlType, argAttrs = [], parentName, e
 
     // Do not convert buffer source types as the impl code can either "get a reference" or "get a copy" to the bytes.
     if (union.ArrayBuffer || union.object) {
-      output.push(`if (${name} instanceof ArrayBuffer) {}`);
+      output.push(`if (utils.isArrayBuffer(${name})) {}`);
     }
     if (union.ArrayBufferViews.size > 0 || union.object) {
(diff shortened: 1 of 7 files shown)


domenic:
This also stops treating SharedArrayBuffer as an ArrayBuffer since that would be incorrect.

lib/types.js
```diff
@@ -195,13 +195,13 @@ function generateTypeConversion(ctx, name, idlType, argAttrs = [], parentName, e
...
     if (union.ArrayBuffer || union.object) {
-      output.push(`if (${name} instanceof ArrayBuffer) {}`);
+      output.push(`if (utils.isArrayBuffer(${name})) {}`);
     }
     if (union.ArrayBufferViews.size > 0 || union.object) {
       let condition = `ArrayBuffer.isView(${name})`;
       // Skip specific type check if all ArrayBufferView member types are allowed.
       if (union.ArrayBufferViews.size !== arrayBufferViewTypes.size) {
-        const exprs = [...union.ArrayBufferViews].map(a => `${name} instanceof ${a}`);
+        const exprs = [...union.ArrayBufferViews].map(a => `${name}.constructor.name === "${a}"`);
```
TimothyGu: BTW another way to do this would be to use `${name}[Symbol.toStringTag]`.

PR #150, opened 2019-12-12
perf(utils): Avoid allocating new exports object

ExE-Boss:
This&nbsp;avoids the&nbsp;overhead of&nbsp;allocating a&nbsp;new&nbsp;`module.exports`&nbsp;object.

I&nbsp;also put&nbsp;the&nbsp;`exports.*` right&nbsp;after each&nbsp;exported&nbsp;member, to&nbsp;make it&nbsp;clearer at&nbsp;a&nbsp;glance what&nbsp;is&nbsp;exported, and&nbsp;what&nbsp;isn’t, without&nbsp;needing to&nbsp;scroll all&nbsp;the&nbsp;way&nbsp;down, and&nbsp;then&nbsp;back&nbsp;up&nbsp;again.

### DIFF
--- a/lib/output/utils.js
+++ b/lib/output/utils.js
@@ -4,15 +4,22 @@
 function isObject(value) {
   return typeof value === "object" && value !== null || typeof value === "function";
 }
+exports.isObject = isObject;
 
 function hasOwn(obj, prop) {
   return Object.prototype.hasOwnProperty.call(obj, prop);
 }
+exports.hasOwn = hasOwn;
 
 const wrapperSymbol = Symbol("wrapper");
+exports.wrapperSymbol = wrapperSymbol;
+
 const implSymbol = Symbol("impl");
+exports.implSymbol = implSymbol;
+
 const sameObjectCaches = Symbol("SameObject caches");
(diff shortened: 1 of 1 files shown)


domenic:
Do you have benchmarks showing that this is a significant improvement in jsdom's performance? Otherwise I don't think we should accept stylistic changes like this.

TimothyGu:
There are some arguments both for and against this. For example, some argue that an object literal helps the engine better figure out the hidden class of the object in one go (and hence provide better performance). Modern JavaScript engines are also able to quite efficiently GC the temporary exports object that was created.

Considering this PR is marked as a performance optimization, but the claim for performance gain is not quite uncontested, I'll move to close this. If you have some benchmark showing this causes a significant performance gain then we could of course reconsider.

ExE-Boss:
The&nbsp;other&nbsp;reason for&nbsp;avoiding re‑assigning&nbsp;`module.exports` is&nbsp;to&nbsp;better support&nbsp;circular&nbsp;imports, which&nbsp;are&nbsp;impossible when&nbsp;`module.exports` is&nbsp;re‑assigned, rather&nbsp;than adding&nbsp;properties to&nbsp;it.

PR #151, opened 2019-12-15
fix: Prevent accidental interface reinstallation

ExE-Boss:
The&nbsp;fact that&nbsp;**WebIDL2JS** allows&nbsp;this is&nbsp;a&nbsp;major&nbsp;footgun, since&nbsp;it&nbsp;creates a&nbsp;new&nbsp;class every&nbsp;time `install`&nbsp;is&nbsp;called, and&nbsp;if&nbsp;it&nbsp;were called&nbsp;twice, it&nbsp;would&nbsp;break the&nbsp;prototype&nbsp;chain of&nbsp;interfaces that&nbsp;were&nbsp;installed in&nbsp;the&nbsp;interim.

---

This&nbsp;has&nbsp;already caused&nbsp;issues for&nbsp;**JSDOM**: [jsdom/jsdom#2753](https://github.com/jsdom/jsdom/pull/2753).

### DIFF
--- a/lib/constructs/interface.js
+++ b/lib/constructs/interface.js
@@ -1424,8 +1424,7 @@ class Interface {
     this.str += `
         if (globalObject[ctorRegistry] === undefined) {
           globalObject[ctorRegistry] = Object.create(null);
-        }
-        if (globalObject[ctorRegistry]["${name}"]) {
+        } else if (globalObject[ctorRegistry]["${name}"]) {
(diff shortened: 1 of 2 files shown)


domenic:
Great catch on the jsdom bug. However, I'm unsure if we should increase the size of every generated file to include this extra check, in order to catch a development-time error. I think that would be better done with a development-time lint or similar. Any ideas there?

PR #152, opened 2019-12-21
fix: Use `util.tryImplForWrapper(…)` for `any` and `object`

### LINKED ISSUE #149: The `any` and `object` types don't generate `utils.tryWrapperForImpl(…)` or `utils.tryImplForWrapper(…)`
This&nbsp;has caused&nbsp;several WPT&nbsp;failures during&nbsp;the&nbsp;development of&nbsp;<https://github.com/jsdom/jsdom/pull/2737>.

Arguably, `object`&nbsp;should&nbsp;require `utils.tryWrapperForImpl` and&nbsp;`utils.tryImplForWrapper` &nbsp;as&nbsp;well.

ExE-Boss:
Fixes #149

---

review?(@domenic, @pmdartus)

### DIFF
--- a/lib/constructs/attribute.js
+++ b/lib/constructs/attribute.js
@@ -22,6 +22,7 @@ class Attribute {
     const sameObject = utils.getExtAttr(this.idl.extAttrs, "SameObject");
 
     const onInstance = utils.isOnInstance(this.idl, this.interface.idl);
+    const idlTypeRaw = this.idl.idlType.idlType;
 
     let objName = `this`;
     if (onInstance) { // we're in a setup method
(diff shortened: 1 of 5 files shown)


TimothyGu:
I'm not quite sure this is a good idea. `object` and `any` values are supposed to be essentially opaque JavaScript object references, and any change to object identity may cause some specs to be unimplementable when using webidl2js. Indeed, C++ browser implementations never unwrap `object` and `any` values, and I don't think we should do anything different here.

On the other hand, I think there's perhaps more value in processing the return value using `tryWrapperForImpl`. But IMHO symmetry is more important here than syntactic sugar.

ExE-Boss:
Well, `obj[@@impl][@@wrapper] === obj`, so&nbsp;object&nbsp;identity will&nbsp;be&nbsp;maintained.

TimothyGu:
Consider the identity (implementation) function. If `obj[@@impl]` is passed to us then we would return `obj` instead of `obj[@@impl]`.

PR #154, opened 2020-01-02
fix: Don’t break on circular imports

ExE-Boss:
Needed&nbsp;for&nbsp;<https://github.com/jsdom/jsdom/pull/2770> and&nbsp;converting&nbsp;`Window` to&nbsp;use&nbsp;**WebIDL** (<https://github.com/jsdom/jsdom/pull/2771>).

Re‑assigning&nbsp;`module.exports` breaks&nbsp;circular&nbsp;`require(…)`s, as&nbsp;any&nbsp;module that&nbsp;was&nbsp;evaluated before&nbsp;`module.exports` was&nbsp;re-assigned will&nbsp;keep the&nbsp;reference to&nbsp;the&nbsp;old&nbsp;object, which&nbsp;is&nbsp;still&nbsp;empty.

By&nbsp;instead creating&nbsp;data&nbsp;properties on&nbsp;`exports` and&nbsp;treating&nbsp;it as&nbsp;a&nbsp;constant, then&nbsp;every&nbsp;reference will&nbsp;have all&nbsp;the&nbsp;properties after&nbsp;all&nbsp;modules have&nbsp;been&nbsp;evaluated.

This&nbsp;is&nbsp;also&nbsp;closer to&nbsp;how&nbsp;**ECMAScript** modules&nbsp;work.

### DIFF
--- a/lib/constructs/attribute.js
+++ b/lib/constructs/attribute.js
@@ -28,7 +28,7 @@ class Attribute {
       objName = `obj`;
     }
     let brandCheck = `
-      if (!this || !module.exports.is(this)) {
+      if (!this || !exports.is(this)) {
         throw new TypeError("Illegal invocation");
       }
     `;
@@ -111,7 +111,7 @@ class Attribute {
 
     if (!this.static && this.idl.special === "stringifier") {
       addMethod("toString", [], `
-        if (!this || !module.exports.is(this)) {
+        if (!this || !exports.is(this)) {
           throw new TypeError("Illegal invocation");
         }
         ${getterBody};
--- a/lib/constructs/dictionary.js
+++ b/lib/constructs/dictionary.js
@@ -70,8 +70,7 @@ class Dictionary {
 
   generate() {
     this.str += `
-      module.exports = {
-        convertInherit(obj, ret, { context = "The provided value" } = {}) {
+      exports.convertInherit = function convertInherit(obj, ret, { context = "The provided value" } = {}) {
     `;
 
     if (this.idl.inheritance) {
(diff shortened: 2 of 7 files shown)


domenic:
As discussed in https://github.com/jsdom/jsdom/pull/2770#issuecomment-570114697 I would prefer not to do this.

ExE-Boss:
Well, this&nbsp;is&nbsp;also&nbsp;closer to&nbsp;how&nbsp;**ECMAScript** modules&nbsp;work.

pmdartus:
Until we drop support for Node 10 (in May 2021) and start leveraging ES module, using `exports.` would greatly help circular dependencies. 

I found myself introducing new circular dependency issues when working https://github.com/jsdom/jsdom/pull/2548. There is one interesting case with `StorageEvent` ([code](https://github.com/jsdom/jsdom/pull/2548/files/f9ff884a8f785856e3f1fa8056b7cce34784bc63#diff-eabcc9e35a8943f69de0558e9fd8fd5fR25)). Adding support for custom elements, forced me to change the modules evaluation order, which had the unexpected effect of exhibiting the circular dependency between `Storage` and `StorageEvent`.  The circular dependency between the interfaces can't be fixed, because the 2 interfaces depend on each other.

> By instead creating data properties on `exports` and treating it as a constant, then every reference will have all the properties after all modules have been evaluated.

Using `exports.` instead of `module.exports` would allow me to move the `require` statement at the top of the file.

domenic:
OK. While I still think that we should avoid circular dependencies whenever possible (e.g. we should merge attributes.js and Attr-impl.js), I can understand that the web platform we're implementing is not so cleanly layered, and so perhaps we should allow them in some cases.

PR #159, opened 2020-01-22
Fix default processor signature

pmdartus:
## Changes

This PR fixes a bug that sneaked in fda810e and that was uncaught by the test suite. If you run the current master version of wedidl2js on the master branch of jsdom it breaks all the tests because the default processor signature is invalid.


test/test.js
```diff
@@ -9,43 +9,63 @@ const casesDir = path.resolve(__dirname, "cases");
...
+      const outputFile = path.resolve(outputDir, path.basename(idlFile, ".webidl") + ".js");
+      const output = fs.readFileSync(outputFile, { encoding: "utf-8" });
+
+      expect(output).toMatchSnapshot();
+    });
+  }
 });
 
-const idlFiles = fs.readdirSync(casesDir);
+describe("with processors", () => {
```
pmdartus: This change doubles the number of tests to validate the transform output with and without processors, which I am not a fan of. It would be preferable to only run the processor tests against the `CEReaction.webidl` and `HTMLConstructor.webidl`, but, it is currently impossible with the available webidl2js API.

domenic: I have always found the current API a bit strange. I think we should probably go this route for now and do a release, but point to https://github.com/jsdom/webidl2js/issues/160 as a potential way of making it better.

I'll hold off on doing a release for now, in case you think we should tackle #160 first.

pmdartus: Agreed, let's move forward with custom elements first. We can always come back to #160 latter.

PR #161, opened 2020-01-29
Move reflection support to host-side using processors

### LINKED ISSUE #15: Support [Reflect] varieties
This would clean up a lot of code in jsdom. Reflect with a default value; reflect URLs; reflect limited to nonnegative integers; ...

### LINKED ISSUE #93: Add a USVString reflector
The behaviour of USVString reflection is defined [here](https://html.spec.whatwg.org/multipage/common-dom-interfaces.html#reflecting-content-attributes-in-idl-attributes%3Aidl-usvstring). jsdom currently implements it as a utility function, [reflectURLAttribute()](https://github.com/tmpvar/jsdom/blob/e2ea031fd763cd65dcd152f915f844a9ad74fb5d/lib/jsdom/utils.js#L130), instead of using webidl2js with `[Reflect]`.

While webidl2js would ideally implement this instead, it requires a URL parser which in turn must have access to the Document and any potential `<base>` elements, thus creating a dependency on [whatwg-url](https://github.com/jsdom/whatwg-url) and access to the DOM. These dependencies are probably too far outside the scope of a general purpose Web IDL to JavaScript converter.

However, the README for `[Reflect]` says that:

> In the future we may move this extended attribute out into some sort of plugin, since it is more related to HTML than to Web IDL.

Doing this should allow us to add a USVString reflector without additional weight or complexity in webidl2js itself.

TimothyGu:
Add proper reflection support through processor

It has long been proposed that reflection features be moved to the host-side, considering it is a part of HTML rather than Web IDL. With the recently added processor infrastructure, this is now possible.

With this change, non-relative imports are now allowed in processors' addImport function. This is done by checking if the imported package fits the npm package name regular expression. Thanks to @ExE-Boss for this contribution.

We also note that existing reflectors did not deal very well with the change from this to esValue, since 2a2a0999a950106094e83ec9d12ea91bad51b0e2. Since this commit moves the reflection code out of webidl2js, the bug now naturally disappears.

Closes #15. Closes #93.


lib/context.js
```diff
@@ -75,10 +81,22 @@ class Context {
     const context = {
       addImport(source, imported) {
         return requires.add(source, imported);
+      },
+      addAbsImport(source, imported) {
```
TimothyGu: This turns out to be necessary for `addImport("whatwg-url")`.

ExE-Boss: I’d&nbsp;prefer if&nbsp;`addImport(…)` supported&nbsp;package&nbsp;imports, which&nbsp;is&nbsp;what I’m&nbsp;doing in&nbsp;<https://github.com/jsdom/webidl2js/pull/162>.

TimothyGu:
Filed https://github.com/w3c/webidl2.js/issues/455 to account for the desire for more flexible extended attributes.

test/reflector.js
```diff
@@ -0,0 +1,52 @@
+"use strict";
```
domenic: I can't see any of this code in the output. Is it just a stub? Maybe we should remove it or make it explicitly a stub?

test/test.js
```diff
@@ -50,6 +51,15 @@ describe("with processors", () => {
         return `
           return ${htmlConstructor}(globalObject, interfaceName);
         `;
+      },
+      processReflect(idl, implObj) {
+        const reflectAttr = idl.extAttrs.find(attr => attr.name === "Reflect");
+        const attrName = reflectAttr.rhs && reflectAttr.rhs.value.replace(/_/g, "-") || idl.name.toLowerCase();
```
domenic: I see this in action in the output, but don't see the effect of the returned object...

test/reflector.js
TimothyGu: You can't "see" it, as the new output is identical to the old output, so there are no diff in the snapshots.

ExE-Boss: The only diff is for tests without processors.

TimothyGu: Ah, yes. That's the confusion.

test/test.js
TimothyGu: See above.

TimothyGu:
Okay, this has been rebased and updated in the following ways:

- `addImport` now supports absolute/package-wide paths.
- Redundant [`Reflect`]'s are now no longer necessary: `[Reflect, ReflectURL]` → `[ReflectURL]`.
- README has been updated.

I'll update jsdom/jsdom#2813 as well soon. PTAL.

README.md
```diff
@@ -128,15 +128,15 @@ By default webidl2js ignores HTML Standard-defined extended attributes [`[CEReac
...
-- `code` is the code generated by webidl2js normally, for calling into the impl class.
+- `code` (string) is the code generated by webidl2js normally, for calling into the impl class.
 
-- `replacementCode` is the new code that will be output in place of `code` in the wrapper class.
+- `replacementCode` (string) is the new code that will be output in place of `code` in the wrapper class.
 
 If either hook is omitted, then the code will not be replaced, i.e. the default is equivalent to `(code) => code`.
 
-Both hooks also have some utility methods that are accessible via `this`:
+Both hooks also has a utility method that is accessible via `this`:
```
domenic:
```suggestion
Both hooks also have a utility method that is accessible via `this`:
```

README.md
```diff
@@ -182,6 +182,69 @@ transformer.generate("wrappers").catch(err => {
...
+An example processor function that implements `boolean` IDL attribute reflection is as follows:
+
+```js
+function processReflect(idl, implName) {
+  // Determine the name of the content attribute being reflected as follows:
+  // - if [Reflect] is provided with a value, then the value is the content attribute.
+  // - otherwise, the content attribute is the name of the IDL attribute, lowercased.
+  const reflectAttr = idl.extAttrs.find(attr => attr.name === "Reflect");
+  const attrName =
+    reflectAttr && reflectAttr.rhs && reflectAttr.rhs.value.replace(/_/g, "-") || idl.name.toLowerCase();
```
domenic: Given that we (or at least I) want to move to strings anyway, we should probably either change or omit this part of the example?

README.md
```diff
@@ -182,6 +182,69 @@ transformer.generate("wrappers").catch(err => {
...
+      set: `
+        if (V) {
+          ${implName}.setAttributeNS(null, "${attrName}", "");
+        } else {
+          ${implName}.removeAttributeNS(null, "${attrName}");
+        }
+      `
+    };
+  }
+  throw new Error(`Unknown IDL type: ${idl.idlType.idlType}`);
```
domenic:
```suggestion
  throw new Error(`Not-yet-implemented IDL type for reflection: ${idl.idlType.idlType}`);
```

README.md
```diff
@@ -414,17 +477,7 @@ Notable missing features include:
...
-### `[Reflect]`
-
-The `[Reflect]` extended attribute is used on IDL attributes to implement the rules for [reflecting a content attribute to an IDL attribute](https://html.spec.whatwg.org/multipage/common-dom-interfaces.html#reflect). If `[Reflect]` is specified, the implementation class does not need to implement any getter or setter logic; webidl2js will take care of it.
-
-By default the attribute passed to `this.getAttribute` and `this.setAttribute` will be the same as the name of the property being reflected. You can use the form `[Reflect=custom]` or `[Reflect=custom_with_dashes]` to change that to be `"custom"` or `"custom-with-dashes"`, respectively.
-
-Note that only the basics of the reflect algorithm are implemented so far: `boolean`, `DOMString`, `long`, and `unsigned long`, with no parametrizations.
-
-In the future we may change this extended attribute to be handled by the caller, similar to `[CEReactions]` and `[HTMLConstructor]`, since it is more related to HTML than to Web IDL.
+A non-standard extended attribute is baked in to webidl2js.
```
domenic:
```suggestion
One non-standard extended attribute is baked in to webidl2js:
```

README.md
TimothyGu: Oops

PR #162, opened 2020-01-30
fix: Allow processors to add package imports

ExE-Boss:
This&nbsp;is&nbsp;needed by&nbsp;<https://github.com/jsdom/webidl2js/pull/161>.

---

See&nbsp;also: <https://github.com/jsdom/webidl2js/pull/163>

PR #158, opened 2020-01-20
Lazy wrapper instantiation

pmdartus:
## Changes

This PR is an early draft for lazy wrapper instantiation. Instead of eagerly create a wrapper for each new impl, the wrapper is only created when accessed by userland code.

While the results are **REALLY** encouraging, there are still a couple of test failures in jsdom (mainly related to events). Running this change against jsdom performance benchmark we can measure a 44% reduction in memory consumption and a 20% to 200% performance improvement across all the tests. More details in the results section.

Fixes #157 

### Memory

| Memory (MB) | p50     | p75     | p90     | p95     | 
|-------------|---------|---------|---------|---------| 
| Baseline    | 1328.97 | 1560.35 | 1669.04 | 1726.82 | 
| Lazy        | 922.80  | 1126.09 | 1315.43 | 1392.72 | 
| Delta       | -44.01% | -38.56% | -26.88% | -23.99% | 

### Benchmark Performance

| Benchmark                                                                | Baseline (ops/sec) | Lazy wrapper (ops/sec) | Detla    | 
|--------------------------------------------------------------------------|--------------------|------------------------|----------| 
| dom/compare-document-position/compare ancestor                           | 18,986.00          | 18,440                 | 2.88%    | 
| dom/compare-document-position/compare descendant                         | 18,110.00          | 19,403                 | -7.14%   | 
| dom/compare-document-position/compare siblings                           | 1,107,701.00       | 1,120,706              | -1.17%   | 
| dom/construction/createComment                                           | 1,276,052.00       | 2,263,382              | -77.37%  | 
| dom/construction/createDocumentFragment                                  | 1,410,648.00       | 2,188,373              | -55.13%  | 
| dom/construction/createElement                                           | 192,790.00         | 646,840                | -235.52% | 
| dom/construction/createEvent                                             | 215,484.00         | 209,256                | 2.89%    | 
| dom/construction/createNodeIterator                                      | 1,217,542.00       | 1,747,421              | -43.52%  | 
| dom/construction/createProcessingInstruction                             | 540,648.00         | 903,657                | -67.14%  | 
| dom/construction/createTextNode                                          | 991,920.00         | 1,732,976              | -74.71%  | 
| dom/inner-html/tables                                                    | 0.31               | 0.39                   | -25.81%  | 
| dom/named-properties/setAttribute(): Remove a named property from window | 3,049.00           | 3,591                  | -17.78%  | 
[...]

PR #156, opened 2020-01-16
perf(proxy): Use per‑class Proxy handler

ExE-Boss:
Currently, a&nbsp;new&nbsp;Proxy&nbsp;handler is&nbsp;created&nbsp;per‑instance, this&nbsp;is&nbsp;inefficient and&nbsp;unnecessary, especially&nbsp;since&nbsp;the&nbsp;proxy&nbsp;handlers don’t&nbsp;depend on&nbsp;any&nbsp;instance‑specific&nbsp;state.

PR #163, opened 2020-01-30
fix: Expose `globalObject` on `proxyHandler` [CEReactions] methods

ExE-Boss:
Needed&nbsp;for&nbsp;<https://github.com/jsdom/jsdom/pull/2548>.

---

See&nbsp;also: <https://github.com/jsdom/webidl2js/pull/162>

### DIFF
--- a/lib/constructs/interface.js
+++ b/lib/constructs/interface.js
@@ -1505,7 +1505,6 @@ class Interface {
 
     this.str += `
       exports.install = function install(globalObject) {
-        const interfaceName = "${name}";
     `;
 
     if (idl.inheritance) {
@@ -1540,6 +1539,9 @@ class Interface {
   }
 
   generate() {
+    this.str += `
+      const interfaceName = "${this.name}";
+    `;
     this.generateIterator();
 
     this.generateExport();
--- a/lib/constructs/interface.js
+++ b/lib/constructs/interface.js
@@ -639,7 +639,7 @@ class Interface {
         `"Failed to set the " + index + " property on '${this.name}': The provided value"`);
       this.requires.merge(conv.requires);
 
-      const prolog = `
+      let prolog = `
         const index = ${P} >>> 0;
         let indexedValue = ${V};
         ${conv.body}
@@ -662,6 +662,8 @@ class Interface {
       }
 
       if (utils.hasCEReactions(this.indexedSetter)) {
+        prolog = `const globalObject = ${O}[impl]._globalObject;` + prolog;
+
         invocation = this.ctx.invokeProcessCEReactions(invocation, {
           requires: this.requires
         });
@@ -678,7 +680,7 @@ class Interface {
         `"Failed to set the '" + ${P} + "' property on '${this.name}': The provided value"`);
       this.requires.merge(conv.requires);
 
-      const prolog = `
+      let prolog = `
         let namedValue = ${V};
         ${conv.body}
       `;
@@ -701,6 +703,9 @@ class Interface {
       }
 
       if (utils.hasCEReactions(this.namedSetter)) {
+        prolog = `
+          const globalObject = ${O}[impl]._globalObject;` + prolog;
+
         invocation = this.ctx.invokeProcessCEReactions(invocation, {
           requires: this.requires
         });
--- a/lib/constructs/interface.js
+++ b/lib/constructs/interface.js
@@ -90,6 +90,7 @@ class Interface {
     this.namedSetter = null;
     this.namedDeleter = null;
     this.stringifier = null;
+    this.needsPerGlobalProxyHandler = false;
 
     this.iterable = null;
     this._analyzed = false;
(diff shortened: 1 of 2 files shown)


domenic:
Hmm, I'm probably missing something obvious, but why does this need to be done here instead of in the jsdom CEReactions processor code?

Also, the test snapshots aren't super-helpful in figuring this out, because they just declare a globalObject variable and then don't use it. Could we add a test that uses it, to illustrate when this is helpful?

Finally, I'll note that there's an alternate approach for getting per-instance state into proxy handlers, which is to make the proxy handler a class (so that methods are shared in the prototype) and then assign any per-instance state as fields on the class. We'd probably want to measure the performance of such an approach though.

ExE-Boss:
The&nbsp;`globalObject` variable&nbsp;is&nbsp;used by&nbsp;the&nbsp;`CEReactions`&nbsp;steps:

https://github.com/jsdom/webidl2js/blob/246f79975a612fd15dff07844a6d9aec2a15135a/test/__snapshots__/test.js.snap#L505-L522

domenic:
Was that an answer to my question of

> Hmm, I'm probably missing something obvious, but why does this need to be done here instead of in the jsdom CEReactions processor code?

or just my question about the test snapshots? If so, I don't quite understand.

ExE-Boss:
After&nbsp;some&nbsp;refactoring, I’ve&nbsp;made&nbsp;it so&nbsp;that&nbsp;there’s now&nbsp;per‑global proxy&nbsp;handlers for&nbsp;interfaces that&nbsp;use&nbsp;`[CEReactions]` on&nbsp;named&nbsp;and&nbsp;indexed setters&nbsp;and&nbsp;deleters.

This&nbsp;keeps the&nbsp;performance&nbsp;gains from&nbsp;#156 for&nbsp;the&nbsp;most&nbsp;common&nbsp;case (no&nbsp;`[CEReactions]`), while&nbsp;ensuring that&nbsp;#130 doesn’t&nbsp;break.

domenic:
Given that [CEReactions] is on the majority of classes in jsdom, and definitely the majority of performance-sensitive ones, I think we should benchmark such a change before committing to it.

pmdartus:
I ran the performance benchmark in jsdom, and I was able to measure a 5-10% performance regressions when introducing this change on certain benchmarks. The benchmark is comparing the current master of jsdom against the custom elements branch + this PR. The performance improvements can probably be attributed to the create element refactor that has been done on the custom element branch.

IMO the performance regression is acceptable compared to value-added by custom element introduction (even if I am completely biased). I am counting a lot on https://github.com/jsdom/webidl2js/pull/158 to improve the overall performance even after the introduction of this regression.

<details>
<summary>Results</summary>

| Test name                                         | Master  (ops/sec) | CE + WeakMap (ops/sec) | Delta   |
|---------------------------------------------------|-------------------|------------------------|---------|
| dom/compare-document-position/compare ancestor    | 20,866            | 16,475                 | -21.04% |
| dom/compare-document-position/compare descendant  | 19,536            | 18,565                 | -4.97%  |
| dom/compare-document-position/compare siblings    | 901,133           | 898,123                | -0.33%  |
| dom/construction/createComment                    | 1,165,566         | 1,301,560              | 11.67%  |
| dom/construction/createDocumentFragment           | 1,320,087         | 1,331,535              | 0.87%   |
| dom/construction/createElement                    | 185,593           | 262,536                | 41.46%  |
| dom/construction/createEvent                      | 189,774           | 210,985                | 11.18%  |
| dom/construction/createNodeIterator               | 1,225,422         | 1,228,464              | 0.25%   |
| dom/construction/createProcessingInstruction      | 648,505           | 645,610                | -0.45%  |
| dom/construction/createTextNode                   | 1,008,861         | 1,037,554              | 2.84%   |
| dom/inner-html/tables                             | 0.31              | 0.29                   | -6.45%  |
| dom/named-properties/setAttribute                 | 3,110             | 3,238                  | 4.12%   |
| dom/tree-modification/appendChild: many parents   | 22.46             | 19.9                   | -11.40% |
| dom/tree-modification/appendChild: many siblings  | 246,965           | 238,447                | -3.45%  |
| dom/tree-modification/appendChild: no siblings    | 244,536           | 212,991                | -12.90% |
| dom/tree-modification/insertBefore: many siblings | 207,314           | 192,440                | -7.17%  |
| dom/tree-modification/removeChild: many parents   | 187,021           | 192,440                | 2.90%   |
| dom/tree-modification/removeChild: many siblings  | 127,226           | 142,754                | 12.21%  |
| dom/tree-modification/removeChild: no siblings    | 106,098           | 115,957                | 9.29%   |
| html/parsing/text                                 | 13.08             | 14.46                  | 10.55%  |
| jsdom/api/new JSDOM() defaults                    | 249               | 261                    | 4.82%   |
| jsdom/api/new JSDOM() with many elements          | 25.36             | 24.38                  | -3.86%  |
</details>

domenic:
I went to merge this but I'm not sure I can write an accurate commit message. Would the following be an accurate summary of the changes?

```
Put globalObject into scope for processCEReactions replacement code

Now the replacement code returned by the processCEReactions handler can reference the globalObject variable, as is documented in the README. This was meant to be introduced in fda810e8d36fe64612bba533228b1b4e91f1a569, but the previous merging of 7c8037ffd4afcff01fd4ac48423675aa64aa1cea prevented that from working properly.
```

I wonder, is `interfaceName` (also documented in the readme) properly in scope still?

ExE-Boss:
Since `interfaceName` is a constant shared across all globals, I’ve moved it into the enclosing module scope.

PR #166, opened 2020-02-13
fix: Add `[ImplicitThis]` behaviour

ExE-Boss:
This&nbsp;implements the&nbsp;`[ImplicitThis]`&nbsp;behaviour present&nbsp;on&nbsp;all WebIDL&nbsp;attributes and&nbsp;operations.

Needed&nbsp;for <https://github.com/jsdom/jsdom/pull/2835>
Will allow fixing <https://github.com/jsdom/jsdom/issues/2830>

See&nbsp;<https://github.com/heycam/webidl/pull/155> for&nbsp;details.

### DIFF
--- a/lib/constructs/attribute.js
+++ b/lib/constructs/attribute.js
@@ -27,7 +27,11 @@ class Attribute {
     if (onInstance) { // we're in a setup method
       objName = `obj`;
     }
-    let brandCheck = `
+    let brandCheck = this.interface.isGlobal ? `
+      if (this !== null && this !== undefined && !exports.is(this)) {
+        throw new TypeError("Illegal invocation");
+      }
+    ` : `
       if (!this || !exports.is(this)) {
         throw new TypeError("Illegal invocation");
       }
--- a/lib/constructs/attribute.js
+++ b/lib/constructs/attribute.js
@@ -118,10 +118,12 @@ class Attribute {
 
     if (!this.static && this.idl.special === "stringifier") {
       addMethod("toString", [], `
-        if (!this || !exports.is(this)) {
+        const esValue = this;
+        if (!exports.is(esValue)) {
           throw new TypeError("Illegal invocation");
         }
-        ${getterBody};
+
+        ${getterBody}
       `, "regular", { configurable, writable: configurable });
     }
 
--- a/lib/constructs/attribute.js
+++ b/lib/constructs/attribute.js
@@ -23,23 +23,15 @@ class Attribute {
 
     const onInstance = utils.isOnInstance(this.idl, this.interface.idl);
 
-    let objName = `this`;
-    if (onInstance) { // we're in a setup method
-      objName = `obj`;
-    }
-    let brandCheck = this.interface.isGlobal ? `
-      if (this !== null && this !== undefined && !exports.is(this)) {
-        throw new TypeError("Illegal invocation");
-      }
-    ` : `
-      if (!this || !exports.is(this)) {
+    let brandCheck = `
+      if (!exports.is(esValue)) {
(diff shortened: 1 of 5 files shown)


domenic:
> [ImplicitThis] 

I think this phrasing is confusing since there is no extended attribute [ImplicitThis].

> Fixes jsdom/jsdom#2830

This is not correct, right? I.e. merging this would not fix that bug.

---

As for the approach, I wish there were a way to generate this only for the cases where it matters, i.e. functions on the prototype chain of the global object. From what I understand that's more of how browser implementations do it. But this works I guess, and the generated code isn't too heavyweight. And there is some value in following the spec as close as possible. So I'm inclined to go with this approach, although it'd be good to get other maintainers' thoughts.

ExE-Boss:
This is how <https://github.com/heycam/webidl/pull/155> is specified to behave.

Case in point: <https://heycam.github.io/webidl/#dfn-create-operation-function>

domenic:
Yes, as I said above,

> I wish there were a way to generate this only for the cases where it matters, i.e. functions on the prototype chain of the global object. From what I understand that's more of how browser implementations do it. But this works I guess, and the generated code isn't too heavyweight. And there is some value in following the spec as close as possible

ExE-Boss:
Right, I somehow missed the paragraph after the horizontal line.

As for <https://github.com/jsdom/jsdom/issues/2830>, that will be fully fixed once this is released and <https://github.com/jsdom/jsdom/pull/2835> is merged.

domenic:
I guess we should be sure not to merge this until Window.js has appropriate brand checks on all its methods; otherwise it would be a regression.

domenic:
Wait, no, that's wrong, I was confused.

Is there any reason to hold off on merging this? I guess maybe because it adds to the output code size slightly for no observable gain at this point...

ExE-Boss:
It removes the need to use `this.addEventListener = this.addEventListener.bind(this)` hack in [`Window.js`](https://github.com/jsdom/jsdom/blob/466856c42da7ba7ba41cc619371c90ffe8b64e48/lib/jsdom/browser/Window.js#L329-L333):
```js
///// METHODS for [ImplicitThis] hack
// See https://lists.w3.org/Archives/Public/public-script-coord/2015JanMar/0109.html
this.addEventListener = this.addEventListener.bind(this);
this.removeEventListener = this.removeEventListener.bind(this);
this.dispatchEvent = this.dispatchEvent.bind(this);
```

lib/constructs/attribute.js
```diff
@@ -23,19 +23,15 @@ class Attribute {
 
     const onInstance = utils.isOnInstance(this.idl, this.interface.idl);
 
-    let objName = `this`;
-    if (onInstance) { // we're in a setup method
-      objName = `obj`;
-    }
```
TimothyGu: Can you tell me if my understanding is correct: this isn't needed because for unforgeable methods/properties `this` always points to `obj` anyway?

ExE-Boss: Well, yes, unless you were to use `unforgeableMethod.call`, in which case `this` will be whatever was passed to `call` as the `thisArg` parameter.

lib/constructs/attribute.js
```diff
@@ -118,10 +118,12 @@ class Attribute {
 
     if (!this.static && this.idl.special === "stringifier") {
       addMethod("toString", [], `
-        if (!this || !exports.is(this)) {
+        const esValue = this;
```
ExE-Boss: This is correct per spec: <https://heycam.github.io/webidl/#es-stringifier>, <https://github.com/heycam/webidl/issues/855>.

PR #171, opened 2020-02-19
fix(interface): Remove dead code

ExE-Boss:
This removes the last of the old‑style mixin dead code that was not removed in <https://github.com/jsdom/webidl2js/pull/135>.

### DIFF
--- a/lib/constructs/interface.js
+++ b/lib/constructs/interface.js
@@ -74,7 +74,6 @@ class Interface {
     this.str = null;
     this.opts = opts;
     this.requires = new utils.RequiresMap(ctx);
-    this.implemented = [];
     this.included = [];
 
     this.constructorOperations = [];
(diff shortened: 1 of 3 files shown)


lib/constructs/interface.js
```diff
@@ -348,7 +346,7 @@ class Interface {
         this.stringifier = member;
       }
     }
-    for (const member of this.inheritedMembers(seen)) {
+    for (const member of this.inheritedMembers()) {
```
TimothyGu: `inheritedMembers` still takes a `seen` parameter (for good reasons, I assume).

ExE-Boss: Well, it does create its own `Set`.

The main reason it takes one is so that **WebIDL2JS** could detect if an interface was present in both the prototype chain and as a legacy mixin.

Since `this.members` no longer takes a `seen` argument, there isn’t any more reason to create the `seen` set ourselves.

TimothyGu: Aha, I missed that it had a default parameter. Looks good then.

TimothyGu:
Could you also update the README? It still has a few references to “mixin interfaces” or #55 that were things of the past.

PR #172, opened 2020-03-07
Implement support for callback interfaces

ExE-Boss:
This adds support for [callback interfaces](https://heycam.github.io/webidl/#idl-callback-interfaces).

Needed for <https://github.com/jsdom/jsdom/issues/2869> (<https://github.com/jsdom/jsdom/pull/2884>).

## To do:
- [x] Update README

---

Supersedes and closes <https://github.com/jsdom/webidl2js/pull/178>.

review?(@domenic, @TimothyGu): I’ve now rebased this on top of <https://github.com/jsdom/webidl2js/pull/182>.


test/__snapshots__/test.js.snap
```diff
@@ -1,5 +1,75 @@
...
+      }
+    }
+  };
+
+  Object.defineProperties(wrapper, {
+    [utils.wrapperSymbol]: { value },
+    value: { value, enumerable: true }
+  });
+
+  return wrapper;
```
domenic: Is wrapper the right name? Aren't we converting the incoming JS value into an internal representation for jsdom to use, not a jsdom-user exposed wrapper?

test/__snapshots__/test.js.snap
```diff
@@ -1,5 +1,75 @@
...
+  }
+
+  const { wrapper } = {
+    wrapper(...args) {
+      let completion;
+      let thisArg = this;
+      let O = value;
+      let X = O;
+
+      try {
```
domenic: Interesting strategy. You're essentially inlining https://heycam.github.io/webidl/#call-a-user-objects-operation into the conversion process, so after we go through the Web IDL bindings, the callback function is essentially represented by a specific "call the user's operation" function for this particular callback. This doesn't follow the spec, but is probably easier for jsdom to deal with. Still, I wonder if there's a way we could make this more obvious. Maybe just documentation...

test/__snapshots__/test.js.snap
```diff
@@ -1,5 +1,75 @@
...
+const interfaceName = \\"AsyncCallbackInterface\\";
+
+exports.convert = function convert(value, { context = \\"The provided value\\" } = {}) {
+  if (!utils.isObject(value)) {
+    throw new TypeError(\`\${context} is not an object.\`);
+  }
+
+  const { wrapper } = {
+    wrapper(...args) {
+      let completion;
```
domenic: This variable appears to be unused.

test/__snapshots__/test.js.snap
```diff
@@ -1,5 +1,75 @@
...
+          if (typeof X !== \\"function\\") {
+            throw new TypeError(\`\${context} does not correctly implement AsyncCallbackInterface.\`);
+          }
+          thisArg = O;
+        }
+
+        for (let i = 0; i < args.length; i++) {
+          args[i] = utils.tryWrapperForImpl(args[i]);
+        }
+        if (args.length < 0) {
```
domenic: This generated code seems pretty bad, especially in the case of 0 args. In this particular case the whole thing could be omitted, including the `...args`, and you could just do `Reflect.apply(X, thisArg, [])`.

I think the guards should be unnecessary since we can always count on jsdom to call with the correct number of args. So we can just have the loop. (And we could even eliminate that, as well as the `...args` parameter declaration itself, when we know `args.length` is `0`.)

test/__snapshots__/test.js.snap
```diff
@@ -1,5 +1,75 @@
...
+
+        return callResult;
+      } catch (err) {
+        return Promise.reject(err);
+      }
+    }
+  };
+
+  Object.defineProperties(wrapper, {
+    [utils.wrapperSymbol]: { value },
```
domenic: Why do we need a wrapper symbol?

test/__snapshots__/test.js.snap
```diff
@@ -1587,46 +1907,141 @@ const Impl = require(\\"../implementations/MixedIn.js\\");
...
+    SHOW_DOCUMENT: { value: 0x100, enumerable: true },
+    SHOW_DOCUMENT_TYPE: { value: 0x200, enumerable: true },
+    SHOW_DOCUMENT_FRAGMENT: { value: 0x400, enumerable: true },
+    SHOW_NOTATION: { value: 0x800, enumerable: true }
+  });
+
+  if (globalObject[ctorRegistry] === undefined) {
+    globalObject[ctorRegistry] = Object.create(null);
+  }
+  globalObject[ctorRegistry][interfaceName] = NodeFilter;
```
domenic: Does it belong in the ctorRegistry? Maybe ctorRegistry is misnamed?

test/__snapshots__/test.js.snap
```diff
@@ -1,5 +1,75 @@
...
+
+const interfaceName = \\"AsyncCallbackInterface\\";
+
+exports.convert = function convert(value, { context = \\"The provided value\\" } = {}) {
+  if (!utils.isObject(value)) {
+    throw new TypeError(\`\${context} is not an object.\`);
+  }
+
+  const { wrapper } = {
+    wrapper(...args) {
```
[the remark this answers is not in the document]
ExE-Boss: This is to ensure that the wrapper can’t be called with `new` but still have a `this  binding.

test/__snapshots__/test.js.snap
ExE-Boss: This is to ensure that the wrapper function doesn’t leak to user code.

test/__snapshots__/test.js.snap
ExE-Boss: I used the same name as @TimothyGu in <https://github.com/jsdom/webidl2js/pull/123>

test/__snapshots__/test.js.snap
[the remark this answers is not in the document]
domenic: Since the callers are all jsdom code, protecting against being called with new isn't necessary, so favor code clarity instead.

test/__snapshots__/test.js.snap
domenic: > This is to ensure that the wrapper function doesn’t leak to user code.

Can you explain how that would happen if the wrapper symbol was omitted?

test/__snapshots__/test.js.snap
domenic: Hmm, well, it seems like the wrong name, so since this is closer to landing, let's change it. "converted" would be OK, or "webIDLRepresentation" maybe, or something like that...

test/__snapshots__/test.js.snap
domenic: One strategy might be, instead of converting to a function with a value property, convert to something like `{ value, callTheUserObjectsOperation }`.

test/__snapshots__/test.js.snap
```diff
@@ -1,5 +1,75 @@
...
+        return callResult;
+      } catch (err) {
+        return Promise.reject(err);
+      }
+    }
+  };
+
+  Object.defineProperties(wrapper, {
+    [utils.wrapperSymbol]: { value },
+    value: { value, enumerable: true }
```
domenic: So the use case for value is for things like https://dom.spec.whatwg.org/#add-an-event-listener's "callback is listener’s callback". However, per https://github.com/whatwg/dom/issues/842 I think this should really be comparing the "object reference".

What do you think of calling this `objectReference` instead, to be closer to the spec? Then, if https://github.com/whatwg/dom/issues/842 gets fixed, it will be more obvious how we're matching it.

Alternately we could advocate for changing the spec's concept name from "object reference" to something nicer (perhaps even "value").

ExE-Boss: If I switch this to use `WeakMap`s, then calling `convert` on the same value multiple times will result in identical wrapper functions.

test/__snapshots__/test.js.snap
ExE-Boss: Probably not.

test/__snapshots__/test.js.snap
ExE-Boss: That’s a leftover from when I was comparing different strategies on how to handle callback interfaces that return `Promise`s.

test/__snapshots__/test.js.snap
ExE-Boss: I went with `internalWrapper` before seeing this.

test/__snapshots__/test.js.snap
ExE-Boss: Well, if there’s a function like:
```webidl
partial interface EventTarget {
	sequence<EventListener> getEventListeners(DOMString type);
};
```

Then the wrapper symbol handles the conversion back to the user provided object is handled by **WebIDL2JS** instead of by **JSDOM**.

There’s also `NodeIterator` and `TreeWalker` that expose a `NodeFilter` attribute.

test/__snapshots__/test.js.snap
ExE-Boss: I just copy and pasted it from <https://github.com/jsdom/webidl2js/pull/123>.

test/__snapshots__/test.js.snap
domenic: That seems incorrect. Per spec a new converted (not wrapper) IDL value is created each time. We should follow that and not add complex caching that isn't needed.

domenic:
I cannot push my fixes to this branch. It would be really nice if you could avoid this separate-organization business and just use personal forks, if you are going to contribute to jsdom.

test/__snapshots__/test.js.snap
domenic: I fixed this in my local copy.

test/__snapshots__/test.js.snap
domenic: Given the documentation, I'm OK with the current version of a function with an objectReference property. In my local version I renamed `converted` to `callTheUserObjectsOperation` to be even clearer.

ExE-Boss:
It says you still haven’t accepted the invite: https://github.com/EB-Forks/jsdom/invitations

domenic:
The invitation has expired.

Regardless, I don't want to have to join another organization to collaborate on my own projects.

lib/types.js
```diff
@@ -117,7 +117,7 @@ function generateTypeConversion(ctx, name, idlType, argAttrs = [], parentName, e
     // string or number type compatible with webidl-conversions
     generateGeneric(`conversions["${idlType.idlType}"]`);
   } else if (resolvedTypes.has(ctx.typeOf(idlType.idlType))) {
-    // dictionaries, enumerations, and interfaces
+    // callbacks, dictionaries, enumerations, and interfaces
```
TimothyGu, line 120:
```suggestion
    // callback interfaces, dictionaries, enumerations, and interfaces
```

You should also delete the `ctx.typeOf(idlType.idlType) === "callback interface"` block that comes afterwards.

lib/types.js
```diff
@@ -477,7 +477,7 @@ function extractUnionInfo(ctx, idlType, errPrefix) {
       seen.object = true;
     } else if (item.idlType === "boolean") {
       seen.boolean = item;
-    } else if (item.idlType === "Function") {
+    } else if (item.idlType === "Function" || ctx.callbackInterfaces.has(item.idlType)) {
```
TimothyGu, line 480: This should be reverted. Callback interfaces are handled below. (Also, callback interfaces and callback functions have different rules for distinguishability.)

README.md
```diff
@@ -286,6 +286,20 @@ This function is mostly used internally, and almost never should be called by yo
 
 jsdom does this for `Window`, which is written in custom, non-webidl2js-generated code, but inherits from `EventTarget`, which is generated by webidl2js.
 
+### For callback interfaces
+
+#### `convert(value, { context })`
+
+Performs the Web IDL conversion algorithm for this callback interface, converting _value_ into a function that performs [call a user object’s operation](https://heycam.github.io/webidl/#call-a-user-objects-operation) when called.
```
TimothyGu, line 293: What's the signature of the returned function? In particular, is there a parameter for setting the `this` value or is the `this` value of the returned function used?

README.md
```diff
@@ -286,6 +286,20 @@ This function is mostly used internally, and almost never should be called by yo
...
+
+The resulting function has an `objectReference` property, which can be used to perform identity checks, as `convert` returns a new function object every time.
```
TimothyGu, line 295: What does `objectReference` refer to? I assume the original `value`. Would be nice to say it.

lib/constructs/callback-interface.js
```diff
@@ -0,0 +1,232 @@
...
+    this._outputStaticProperties = new Map();
+  }
+
+  _analyzeMembers() {
+    for (const member of this.idl.members) {
+      switch (member.type) {
+        case "operation":
+          if (this.operation !== null) {
+            throw new Error(
+              `Callback interface ${this.name} has more than one operation`
```
TimothyGu, line 29: Nothing seems to prevent callback interfaces from having more than one operation. See e.g., this test case here: https://source.chromium.org/chromium/chromium/src/+/master:third_party/blink/renderer/bindings/tests/idls/core/test_callback_interface.idl;l=30;drc=0aee4434a4dba42a42abaea9bfbc0cd196a63bc1. But on the other hand, I can't actually find a spec that puts more than two operations into the same callback interface. Hmm.

We should probably file a spec bug.

TimothyGu: Wait, never mind.

> Callback interfaces must define exactly one regular operation.

lib/constructs/callback-interface.js
```diff
@@ -0,0 +1,232 @@
...
+      this.requires.merge(data.requires);
+    }
+  }
+
+  addStaticProperty(name, body, { configurable = true, enumerable = typeof name === "string", writable = true } = {}) {
+    const descriptor = { configurable, enumerable, writable };
+    this._outputStaticProperties.set(name, { body, descriptor });
+  }
+
+  addProperty() {}
```
TimothyGu, line 63: Could you add a comment saying why this is empty? 1) We have to define this function because other related classes like `Constant` use it, and 2) because callback interfaces don't generate platform objects.

lib/constructs/callback-interface.js
```diff
@@ -0,0 +1,232 @@
...
+              `Callback interface ${this.name} has more than one operation`
+            );
+          }
+          this.operation = member;
+          break;
+        case "const":
+          this.constants.set(member.name, new Constant(this.ctx, this, member));
+          break;
+        default:
+          if (!this.ctx.options.suppressErrors) {
```
TimothyGu, line 38: We probably don't need to guard this check. It's only used for things like unrecognized types.

lib/constructs/callback-interface.js
```diff
@@ -0,0 +1,232 @@
...
+  }
+
+  addProperty() {}
+
+  generateConversion() {
+    const { operation, name } = this;
+    const opName = operation.name;
+
+    const isAsyncInUnion = operation.idlType.union &&
+                           operation.idlType.idlType.some(idlType => idlType.generic === "Promise");
```
TimothyGu, line 70: The spec just checks if the return type is a promise type. No need to deal with unions. (A promise type is never distinguishable with any other type, so it can't appear in a union anyway.)

lib/constructs/callback-interface.js
```diff
@@ -0,0 +1,232 @@
...
+              X = O[${utils.stringifyPropertyName(opName)}];
+              if (typeof X !== "function") {
+                throw new TypeError(\`\${context} does not correctly implement ${name}.\`)
+              }
+              thisArg = O;
+            }
+    `;
+
+    // We don't implement all of https://heycam.github.io/webidl/#web-idl-arguments-list-converting since the callers
+    // are assumed to always pass the correct number of arguments and we don't support optional/variadic arguments.
```
TimothyGu, line 107: Maybe link to https://github.com/jsdom/webidl2js/issues/71 here?

README.md
```diff
@@ -286,6 +286,20 @@ This function is mostly used internally, and almost never should be called by yo
 
 jsdom does this for `Window`, which is written in custom, non-webidl2js-generated code, but inherits from `EventTarget`, which is generated by webidl2js.
 
+### For callback interfaces
+
+#### `convert(value, { context })`
+
+Performs the Web IDL conversion algorithm for this callback interface, converting _value_ into a function that performs [call a user object’s operation](https://heycam.github.io/webidl/#call-a-user-objects-operation) when called, with `thisArg` being the `this` value of the converted function.
+
+The resulting function has an `objectReference` property, which is the same object as `value` and can be used to perform identity checks, as `convert` returns a new function object every time.
```
TimothyGu, line 295: Also mention that we don't implement the [callback context](https://heycam.github.io/webidl/#dfn-callback-context) yet.

lib/constructs/callback-interface.js
ExE-Boss: I copied this from `interface.js`, which has this check: <https://github.com/jsdom/webidl2js/blob/6b2784221206174e745e2756e7ed2f4f8d7bb75c/lib/constructs/interface.js#L301-L304>

README.md
ExE-Boss: The callback context is used to ensure that the callbacks are executed in the correct realm, which is handled for us by the platform.

lib/constructs/callback-interface.js
TimothyGu: Heh. I think that might be because there are things allowed in interfaces that we don’t support yet, like async iterables. In contrast, anything else here would be in violation of the Web IDL spec.

TimothyGu:
Thanks!

PR #174, opened 2020-03-09
fix: Make constructor registry properties immutable

ExE-Boss:
This makes it so that the constructor registry is an immutable property and registered constructors are also immutable properties.

This prevents accidentally overwriting registered constructors or deleting the registry.

The registry is also made non-enumerable, so that it doesn’t show by default in the output of [Node’s `util.inspect(…)` function](https://nodejs.org/api/util.html#util_util_inspect_object_options).

### DIFF
--- a/lib/constructs/interface.js
+++ b/lib/constructs/interface.js
@@ -1458,6 +1458,14 @@ class Interface {
       `;
     }
 
+    this.str += `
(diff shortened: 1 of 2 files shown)


domenic:
This seems like extra complexity for no gain. This kind of "defensive coding" is not necessary in this project.

PR #178, opened 2020-03-14
Implement callback interfaces

domenic:
Needed for fixing https://github.com/jsdom/jsdom/issues/2869 (used in https://github.com/jsdom/jsdom/pull/2884).

Closes #172 by superceding it. I can actually edit this branch.


domenic:
The code for generating the constants seems really over-complicated, but I guess it is basically a copy and paste of the code for interfaces. I spent five minutes trying to simplify but it got kind of messy, so I won't pursue that for now. If people have ideas though, I'd be interested.

lib/constructs/callback-interface.js
```diff
@@ -0,0 +1,233 @@
...
+    // are assumed to always pass the correct number of arguments and we don't support optional/variadic arguments.
+    for (const argName of argNames) {
+      this.str += `
+            ${argName} = utils.tryWrapperForImpl(${argName});
+      `;
+    }
+
+    const argsToReflectCall = argNames.length > 0 ? `, ${argNames.join(", ")}` : "";
+    this.str += `
+            let callResult = Reflect.call(X, thisArg${argsToReflectCall});
```
ExE-Boss:
```suggestion
    this.str += `
            let callResult = Reflect.apply(X, thisArg, [${argNames.join(", ")}]);
```

lib/types.js
```diff
@@ -26,12 +26,7 @@ function mergeExtAttrs(a = [], b = []) {
 }
 
 // Types of types that generate an output file.
-const resolvedTypes = new Set([
-  "callback interface",
-  "dictionary",
-  "enumeration",
-  "interface"
-]);
+const resolvedTypes = new Set(["callback interface", "dictionary", "enumeration", "interface"]);
```
ExE-Boss: I made this multi‑line so that when there’s multiple in‑flight PRs that add new types (like #123), they avoid conflicting with each other.

domenic: I don't think that's a sufficient motivation for making the code more verbose in that way. In general, please do not change project style without opening an issue for discussion.

lib/constructs/callback-interface.js
ExE-Boss: There is no such thing as `Reflect.call(…)`, so this line will throw a `TypeError`.

I’ve fixed that in <https://github.com/jsdom/webidl2js/pull/172>.

PR #179, opened 2020-03-16
refactor: Rename imported symbols

ExE-Boss:
I’ve waited with opening this until after <https://github.com/jsdom/webidl2js/pull/166> was done to avoid merge conflicts.

This frees up the `impl` variable for use in <https://github.com/jsdom/webidl2js/pull/158> and also makes it clearer that the current `impl` variable refers to `implSymbol`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,8 +26,7 @@ will generate a JavaScript wrapper class file roughly like this:
 
 ```js
 const conversions = require("webidl-conversions");
-const impl = require("./utils.js").implSymbol;
-const ctorRegistry = require("./utils.js").ctorRegistrySymbol;
+const { implSymbol, ctorRegistrySymbol } = require("./utils.js");
 
 const Impl = require("./SomeInterface-impl.js").implementation;
 
@@ -56,7 +55,7 @@ class SomeInterface {
       context: "Failed to execute 'add' on 'SomeInterface': parameter 2"
     });
 
-    return this[impl].add(...args);
+    return this[implSymbol].add(...args);
   }
 }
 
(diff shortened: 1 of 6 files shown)


lib/constructs/interface.js
```diff
@@ -492,8 +492,7 @@ class Interface {
   }
 
   generateRequires() {
-    this.requires.addRaw("impl", "utils.implSymbol");
-    this.requires.addRaw("ctorRegistry", "utils.ctorRegistrySymbol");
+    this.requires.addRaw(`{ implSymbol, ctorRegistrySymbol }`, "utils");
```
ExE-Boss: I’ve changed this to use object destructuring to match how alias imports are done in **JSDOM**.

README.md
```diff
@@ -26,8 +26,9 @@ will generate a JavaScript wrapper class file roughly like this:
 
 ```js
 const conversions = require("webidl-conversions");
-const impl = require("./utils.js").implSymbol;
-const ctorRegistry = require("./utils.js").ctorRegistrySymbol;
+const utils = require("./utils.js");
+
+const { implSymbol, ctorRegistrySymbol } = utils;
```
TimothyGu: `utils` doesn't seem to be used anywhere. (This is an example so we don't have to do the exact same as the output.)

```suggestion
const { implSymbol, ctorRegistrySymbol } = require("./utils.js");
```

TimothyGu:
Rebased and fixed some issues that came up in rebasing (96794fa).

ExE-Boss:
I had originally fixed that as part of <https://github.com/jsdom/webidl2js/commit/42e8cd58691c203d62ac79b6a39dffd7ff86bac1>.

TimothyGu:
It’s possible that rebasing destroyed the merge. For future could you please rebased rather than merge

PR #182, opened 2020-03-25
Preliminary support for callback interfaces

TimothyGu:
Implemented JS-to-IDL type conversion but not:

- Saving [callback context][1]
- [Call a user object's operation][2]
- IDL-to-JS type conversion (#71)
- [Legacy callback interface object][3]

Those things are left to more ambitious pull requests in the future.

Fixes https://github.com/jsdom/jsdom/issues/2869.

[1]: https://heycam.github.io/webidl/#dfn-callback-context
[2]: https://heycam.github.io/webidl/#call-a-user-objects-operation
[3]: https://heycam.github.io/webidl/#dfn-legacy-callback-interface-object

----

I've tested this by linking webidl2js in jsdom (also https://github.com/jsdom/jsdom/pull/2919) and it does fix https://github.com/jsdom/jsdom/issues/2869. Some things in this PR are gathered from elsewhere (like my own https://github.com/jsdom/webidl2js/pull/123). But others, like union extraction and type distinguishability bits, were written anew, as they are not handled correctly by any of the other open PRs on callback interfaces (#123, #172, #178). This PR should complement with them nicely.

### DIFF
--- a/test/__snapshots__/test.js.snap
+++ b/test/__snapshots__/test.js.snap
@@ -1112,6 +1112,128 @@ const Impl = require(\\"../implementations/Enum.js\\");
 "
 `;
 
+exports[`with processors EventTarget.webidl 1`] = `
+"\\"use strict\\";
+
+const conversions = require(\\"webidl-conversions\\");
+const utils = require(\\"./utils.js\\");
+
+const implSymbol = utils.implSymbol;
+const ctorRegistrySymbol = utils.ctorRegistrySymbol;
+
+const interfaceName = \\"EventTarget\\";
+
+exports.is = function is(obj) {
+  return utils.isObject(obj) && utils.hasOwn(obj, implSymbol) && obj[implSymbol] instanceof Impl.implementation;
+};
+exports.isImpl = function isImpl(obj) {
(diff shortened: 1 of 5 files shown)


test/__snapshots__/test.js.snap
```diff
@@ -1112,6 +1112,128 @@ const Impl = require(\\"../implementations/Enum.js\\");
...
+        args.push(curArg);
+      }
+      {
+        let curArg = arguments[1];
+        if (curArg === null || curArg === undefined) {
+          curArg = null;
+        } else {
+          if (!utils.isObject(curArg)) {
+            throw new TypeError(
+              \\"Failed to execute 'addEventListener' on 'EventTarget': parameter 2\\" + \\" is not an object\\"
```
domenic: I think you could make this all one string by tweaking the source.

TimothyGu: I'd like to punt on this. This is slightly difficult, considering I'm currently doing

```js
`${errPrefix} + " dictionary"`
```

and `errPrefix` can really be anything, string literal or not. This is also not the only place in the snapshot that has something like this.

TimothyGu:
> So #178 or similar would be done on top of this?

Yes.

PR #43, opened 2017-05-15
Add support for typedef and unions

TimothyGu:

PR #183, opened 2020-03-26
refactor(utils): Avoid re‑assigning `module.exports`

ExE-Boss:
This refactors `utils.js` to create properties on `exports`, rather than re‑assigning `module.exports`, which matches the rest of the generated files and all util files in **JSDOM**.

---

review?(@pmdartus)

### DIFF
--- a/lib/output/utils.js
+++ b/lib/output/utils.js
@@ -1,18 +1,21 @@
 "use strict";
 
 // Returns "Type(value) is Object" in ES terminology.
-function isObject(value) {
+exports.isObject = value => {
   return typeof value === "object" && value !== null || typeof value === "function";
-}
+};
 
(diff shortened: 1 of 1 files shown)


TimothyGu:
Please don't reformat existing files for no particular gains.

ExE-Boss:
The gain is that this makes it clearer which exports are used.

Incidentally, why was <https://github.com/jsdom/webidl2js/pull/43> allowed to reformat `utils.js`?

TimothyGu:
That was quite a few years ago, let's see…

One answer is because the new functions I defined referenced each other but also needed to be exported, and I'm not too big of a fan of simultaneous assignments in the `const abc = exports.abc = () => …`. That's in comparison to this PR, which is all style.

Another answer is because I was young and didn't know better ;) I have been convinced by numerous style guides that I've read in the past three years that the tradeoffs for deliberate reformatting are not good.

PR #190, opened 2020-04-03
Add `webidl2js‑globals.js` to handle setup of [Global]s

ExE-Boss:
## TODO:

- [ ] Fix the interface order determination algorithm so that `Element` goes after `Node` and `EventTarget`, but before `HTMLElement`.

---

Supersedes and closes <https://github.com/jsdom/webidl2js/pull/101>.

review?(@TimothyGu)


TimothyGu:
Perhaps we can be a bit less ambitious, and separate out the parts that add [Exposed] support to `install()` (which I mentioned in #191) from the parts that aggregates globals into the same file? This way we could immediately start using it in jsdom.

WRT an algorithm to figure out the require order, if the dependency graph turns out to be acyclic then [topological sorting](https://en.wikipedia.org/wiki/Topological_sorting) should work. If it is cyclic then… we’d have to look into other things.

ExE-Boss:
The dependency order is already enforced to be acyclic in: <https://github.com/jsdom/webidl2js/blob/5af205c3dd63e7248ba2e9039b915fa529aa678f/lib/constructs/interface.js#L468-L478>

TimothyGu:
I've rebased this, but it doesn't look like it's ready for review yet? Interface files call `webidl2jsGlobal.setupGlobal` but it's not defined.

ExE-Boss:
That was before I renamed it to just `installInterfaces`, which is more descriptive of what it does.

PR #191, opened 2020-04-03
Implement support for `[LegacyWindowAlias]` and `[NoInterfaceObject]`

ExE-Boss:
Fixes #187

---

review?(@TimothyGu)

PR #192, opened 2020-04-04
Implement support for the `[Exposed]` extended attribute

ExE-Boss:
Extracted from <https://github.com/jsdom/webidl2js/pull/190> at @TimothyGu’s request in <https://github.com/jsdom/webidl2js/pull/191#pullrequestreview-387619767>.

This also fixes a bug where `install` takes `globalName` instead of `globalNames`.

---

This will free up <https://github.com/jsdom/webidl2js/pull/190> to only handle the creation of `webidl2js‑globals.js`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -454,6 +454,7 @@ webidl2js is implementing an ever-growing subset of the Web IDL specification. S
 - Variadic arguments
 - `[Clamp]`
 - `[EnforceRange]`
+- `[Exposed]`
 - `[LegacyArrayClass]`
 - `[LegacyUnenumerableNamedProperties]`
 - `[LegacyWindowAlias]`
--- a/README.md
+++ b/README.md
@@ -264,9 +264,11 @@ Performs the Web IDL conversion algorithm for this interface, converting _value_
 
 In practice, this means doing a type-check equivalent to `is(value)`, and if it passes, returns the corresponding impl. If the type-check fails, it throws an informative exception. _context_ can be used to describe the provided value in any resulting error message.
 
-#### `install(globalObject)`
+#### `install(globalObject, globalNames)`
 
 This method creates a brand new wrapper constructor and prototype and attach it to the passed `globalObject`. It also registers the created constructor with the `globalObject`'s global constructor registry, which makes `create()`, `createImpl()`, and `setup()` work. (Thus, it is important to invoke `install()` before invoking those methods, as otherwise they will throw.)
+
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -454,6 +454,7 @@ webidl2js is implementing an ever-growing subset of the Web IDL specification. S
 - Variadic arguments
 - `[Clamp]`
 - `[EnforceRange]`
+- `[Exposed]`
```
TimothyGu, line 457: Make sure to remove this from the list of unsupported features.

lib/constructs/callback-interface.js
```diff
@@ -178,14 +197,41 @@ class CallbackInterface {
...
+        const exposed = new Set([
+      `;
+
+      for (const globalName of this.exposed) {
+        this.str += `"${globalName}",\n`;
+      }
+
+      this.str += `
+        ]);
+      `;
```
TimothyGu, line 211: What about just `const exposed = new Set(${JSON.stringify([...this.exposed])});`

lib/constructs/callback-interface.js
```diff
@@ -178,14 +197,41 @@ class CallbackInterface {
...
+        for (const globalName of globalNames) {
+          if (exposed.has(globalName)) {
+            isExposed = true;
+            break;
+          }
+        }
+
+        if (!isExposed) {
+          return;
+        }
```
TimothyGu, line 233: Can we simplify it to just `if (!globalNames.some(name => exposed.has(name)) return;`

lib/constructs/interface.js
```diff
@@ -1463,7 +1463,47 @@ class Interface {
...
-      exports.install = (globalObject, globalName) => {
+      const exposed = new Set([
+    `;
+
+    for (const globalName of this.exposed) {
+      this.str += `"${globalName}",\n`;
+    }
+
+    this.str += `
+      ]);
```
TimothyGu, line 1474: Same comments as above.

lib/constructs/interface.js
```diff
@@ -1463,7 +1463,47 @@ class Interface {
...
+        for (const globalName of globalNames) {
+          if (exposed.has(globalName)) {
+            isExposed = true;
+    `;
+
+    if (this.legacyWindowAliases) {
+      this.str += `
+            if (globalName === "Window") {
+              isWindow = true;
+            }
```
TimothyGu, line 1494: Same comments as above. For the `isWindow` check, another `globalNames.some()` should be fine too.

lib/constructs/callback-interface.js
ExE-Boss: `globalNames` is a `Set` in <https://github.com/jsdom/webidl2js/pull/190>.

TimothyGu: Ah, someone should really make a `Set.prototype.some`/`every` proposal to TC39. But I'd rather the API be taking an array.

ExE-Boss: There is the **Iterator Helpers** proposal: <https://tc39.es/proposal-iterator-helpers/>.

PR #194, opened 2020-04-16
Implement support for callback functions

ExE-Boss:
This implements callback functions similarly to how callback interfaces were implemented in <https://github.com/jsdom/webidl2js/pull/172>.

I would’ve preferred if this were included in **WebIDL2JS v16.0**.

## Depends on:

- [ ] <https://github.com/jsdom/webidl2js/pull/219>

---

Supersedes and closes <https://github.com/jsdom/webidl2js/pull/123>

review?(@domenic, @TimothyGu, @Sebmaster)

### DIFF
--- a/lib/constructs/callback-function.js
+++ b/lib/constructs/callback-function.js
@@ -17,13 +17,14 @@ class CallbackFunction {
     this.str = null;
 
     this.requires = new utils.RequiresMap(ctx);
+
+    this.legacyTreatNonObjectAsNull = Boolean(utils.getExtAttr(idl.extAttrs, "LegacyTreatNonObjectAsNull"));
   }
 
   generateConversion() {
-    const { idl } = this;
+    const { idl, legacyTreatNonObjectAsNull } = this;
     const isAsync = idl.idlType.generic === "Promise";
 
-    const legacyTreatNonObjectAsNull = Boolean(utils.getExtAttr(idl.extAttrs, "LegacyTreatNonObjectAsNull"));
     const assertCallable = legacyTreatNonObjectAsNull ? "" : `
       if (typeof value !== "function") {
         throw new TypeError(context + " is not a function");
--- a/lib/constructs/callback-function.js
+++ b/lib/constructs/callback-function.js
@@ -1,5 +1,7 @@
 "use strict";
 
+const conversions = require("webidl-conversions");
+
 const utils = require("../utils.js");
 const Types = require("../types.js");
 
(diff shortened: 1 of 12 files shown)


test/__snapshots__/test.js.snap
```diff
@@ -1,5 +1,81 @@
...
+exports.convert = (value, { context = \\"The provided value\\" } = {}) => {
+  if (typeof value !== \\"function\\") {
+    throw new TypeError(context + \\" is not a function\\");
+  }
+
+  function invokeTheCallbackFunction() {
+    const thisArg = utils.tryWrapperForImpl(this);
+    let callResult;
+
+    callResult = Reflect.apply(value, thisArg, []);
```
domenic, line 64: Optional, but if it's easy, it would be nice to remove the unused `callResult` variable from this case (and construct).

test/__snapshots__/test.js.snap
```diff
@@ -5248,6 +5402,130 @@ const Impl = require(\\"../implementations/URL.js\\");
...
+      args[i] = utils.tryWrapperForImpl(args[i]);
+    }
+
+    if (args.length < 2) {
+      for (let i = args.length; i < 2; i++) {
+        args[i] = undefined;
+      }
+    } else if (args.length > 2) {
+      args.length = 2;
+    }
```
domenic, line 5432: All this args handling seems unnecessary. Can we assume that the caller (which will be jsdom code) will always pass the correct number of arguments, and then just do this similar to what callback interfaces did? So this ends up reducing to

```js
function invokeTheCallbackFunction(url, string) {
  // ...
  url = utils.tryWrapperForImpl(url);
  /* optional, if it's easier to not distinguish interfaces etc.: */ string = utils.tryWrapperForImpl(string);
  // ...
}
```

test/__snapshots__/test.js.snap
```diff
@@ -5248,6 +5402,130 @@ const Impl = require(\\"../implementations/URL.js\\");
...
+
+    callResult = Reflect.apply(value, thisArg, args);
+
+    callResult = URL.convert(callResult, { context: context });
+
+    return callResult;
```
domenic, line 5438: The result should be an impl, since the caller is jsdom code. Right?

ExE-Boss: That’s handled by `URL.convert(…)` above.

test/__snapshots__/test.js.snap
ExE-Boss: The issue is that we have to support optional and variadic arguments in callback functions, because we support APIs that use them.

test/cases/URLCallback.webidl
```diff
@@ -0,0 +1 @@
+callback URLCallback = URL (URL? url, DOMString string);
```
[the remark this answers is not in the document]
ExE-Boss: It doesn’t, but the shapshot tests will end up missing the converted files otherwise.

test/__snapshots__/test.js.snap
domenic: Sure, the same for callback interfaces. But we could check for the simple case in https://github.com/jsdom/webidl2js/pull/172/files#diff-7dc541e84f0e601cb83c2704397cb01aR72-R75

ExE-Boss:
I’ll begin work on the **JSDOM** side of this once <https://github.com/jsdom/jsdom/pull/2884> is merged.

lib/types.js
```diff
@@ -201,11 +215,14 @@ function generateTypeConversion(ctx, name, idlType, argAttrs = [], parentName, e
...
-
-    if (union.sequenceLike || union.dictionary || union.record || union.object || union.callbackInterface) {
+    if (
+      union.sequenceLike ||
+      union.dictionary ||
+      union.record ||
+      union.object ||
+      union.callback ||
+      union.callbackInterface
+    ) {
```
TimothyGu, line 225: Please follow the spec for union type conversion as closely as possible: https://heycam.github.io/webidl/#es-union

lib/types.js
```diff
@@ -402,7 +424,7 @@ function extractUnionInfo(ctx, idlType, errPrefix) {
     numeric: null,
     boolean: null,
     // Callback function, not interface
-    callback: false,
+    callback: null,
```
TimothyGu, line 427: Let's just call this `callbackFunction`.

lib/constructs/callback-function.js
```diff
@@ -0,0 +1,206 @@
...
+  }
+
+  toString() {
+    this.str = "";
+    this.generate();
+    return this.str;
+  }
+}
+
+CallbackFunction.prototype.type = "callback";
```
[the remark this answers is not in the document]
ExE-Boss: This is to match **WebIDL2**’s `type` property: <https://github.com/w3c/webidl2.js/blob/561efd0007038cc94ec2da8fcfa11623fe94101a/lib/productions/callback.js#L22-L24>.

test/__snapshots__/test.js.snap
```diff
@@ -1,5 +1,149 @@
...
+    try {
+      callResult = Reflect.apply(value, thisArg, []);
+
+      callResult = Promise.resolve(callResult).then(
+        value => {
+          value = conversions[\\"any\\"](value, { context: context + \\" promise value\\" });
+
+          return value;
+        },
+        reason => reason
```
domenic, line 115: This is incorrect.

test/__snapshots__/test.js.snap
```diff
@@ -1,5 +1,149 @@
...
+      throw new Error(\\"Internal error: invokeTheCallbackFunction is not a constructor\\");
+    }
+
+    const thisArg = utils.tryWrapperForImpl(this);
+    let callResult;
+
+    try {
+      callResult = Reflect.apply(value, thisArg, []);
+
+      callResult = Promise.resolve(callResult).then(
```
domenic, line 109: This doesn't implement the semantics described in https://heycam.github.io/webidl/#es-invoking-callback-functions + https://heycam.github.io/webidl/#es-promise . For that the correct code would be something more like

```js
return new Promise(resolve => resolve(callResult))
```

Note that `Promise.resolve()` is inappropriate because it returns its argument in some cases, which is never true for converting to a Web IDL promise type. And note how converting to a Web IDL `Promise<T>` type ignores the `T`.

ExE-Boss: This code is generated by: <https://github.com/jsdom/webidl2js/blob/7d45dfc6e6d1ba737b77eaabbab712dcf2d19435/lib/types.js#L330-L349>

Changing that is outside the scope of this PR.

test/__snapshots__/test.js.snap
ExE-Boss: This code is generated by: <https://github.com/jsdom/webidl2js/blob/7d45dfc6e6d1ba737b77eaabbab712dcf2d19435/lib/types.js#L330-L349>

Changing that is outside the scope of this PR.

test/__snapshots__/test.js.snap
domenic: Well, I'm not willing to merge this PR if it generates incorrect code that is unusable for consumers of webidl2js, so I guess fixing that is a dependency of this PR then.

ExE-Boss: Done in <https://github.com/jsdom/webidl2js/pull/219>.

domenic:
I'm unable to push commits to this branch again, despite it seemingly being a normal fork :(. I'll merge directly into master from my local copy.

domenic:
Merged as https://github.com/jsdom/webidl2js/commit/4a6e558c1fddb9ea594241f4025c992d07b2e191!

PR #195, opened 2020-04-17
Remove undocumented Impl.init functionality

domenic:
This is not used by jsdom and seems unnecessary.

### DIFF
--- a/lib/constructs/interface.js
+++ b/lib/constructs/interface.js
@@ -1187,9 +1187,6 @@ class Interface {
 
     this.str += `
         obj[implSymbol][utils.wrapperSymbol] = obj;
-        if (Impl.init) {
-          Impl.init(obj[implSymbol], privateData);
-        }
         return obj;
       };
     `;
--- a/test/__snapshots__/test.js.snap
+++ b/test/__snapshots__/test.js.snap
@@ -105,9 +105,6 @@ exports.setup = function setup(obj, globalObject, constructorArgs = [], privateD
   });
 
   obj[implSymbol][utils.wrapperSymbol] = obj;
-  if (Impl.init) {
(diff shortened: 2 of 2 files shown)


domenic:
I've confirmed that this works on jsdom just as well as the current master does. (The current master seems to have problems but I think they are solved by https://github.com/jsdom/jsdom/pull/2884.)

domenic:
It seems `DOMException`'s usage could be replaced by moving that code into the constructor, do you agree?

```js
exports.implementation = class DOMExceptionImpl {
  constructor(globalObject, [message, name], { wrapper }) {
    this.name = name;
    this.message = message;
    
    if (Error.captureStackTrace) {
      Error.captureStackTrace(wrapper, wrapper.constructor);
    }
  }

  get code() {
    return legacyErrorCodes[this.name] || 0;
  }
};
```

ExE-Boss:
That still won’t support **SpiderMonkey**’s `Error.prototype.stack` getter, which requires the wrapper to be created using `Reflect.construct(Error, args, DOMException)`.

domenic:
I don't understand the relevance of that comment. SpiderMonkey is not supported today and the wrapper is not created in that way.

TimothyGu:
I think the `new` method should also create the stack property. Moving it into the impl constructor would mean that the `stack` property is only added when create/createImpl is called.

domenic:
Hmm, I think you're right. I wish there were a way to do that without adding three lines to every generated output file (possibly 6 if I can't find a good refactoring).

The only thing I can think of is a custom extended attribute, e.g. `[WebIDL2JSHasExtraInit] interface DOMException { ... }`, which is not very fun either.

I guess we can just wait until we get optional chaining syntax and raise the minimum V8 version that high, to bring it down to 1 line.

PR #196, opened 2020-04-17
Add new() export for interfaces

### LINKED ISSUE #193: Separating constructor and new?
I'm trying to implement https://github.com/whatwg/streams/pull/1035 and running into the issue that we don't have a way to invoke the Web IDL [new](https://heycam.github.io/webidl/#new) operation without also invoking the impl constructor.

The impl constructor is what backs the wrapper constructor. But in specs we have two different paths for creating platform objects:

- `new PlatformObject()` from JS code runs [these steps](https://heycam.github.io/webidl/#create-an-interface-object)
- Specs will do "[new](https://heycam.github.io/webidl/#new) `PlatformObject`" and then initialize them directly.

We don't currently have a way of doing the latter.

I'll probably investigate with some patches locally in the streams repo, but this issue can track upstreaming them.

domenic:
Closes #193.

Since I had to do some refactoring to reduce duplication anyway, this also includes general cleanup to the output, such as replacing `obj` with either `value` or `wrapper`, using arrow functions for our exports, and such.

Ideas welcome on further refactoring to reduce duplication, but I think it's probably easiest to wait until we can stop exporting `setup()`; once that's true things become simpler.

On top of #195.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -157,7 +157,7 @@
     "linebreak-style": ["error", "unix"],
     "lines-around-comment": "off",
     "max-depth": "off",
-    "max-len": ["error", 120, { "ignoreUrls": true }],
+    "max-len": ["error", 120, { "ignoreUrls": true, "ignoreTemplateLiterals": true }],
     "max-nested-callbacks": "off",
     "max-params": "off",
     "max-statements": "off",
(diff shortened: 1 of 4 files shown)


domenic:
I've confirmed that this works on jsdom just as well as the current master does. (The current master seems to have problems but I think they are solved by https://github.com/jsdom/jsdom/pull/2884.)

PR #200, opened 2020-04-17
Ensure promise-typed operations and attributes return promises

### LINKED ISSUE #79: Promise-returning operations should return Promise.reject on invalid argument
If an interface defines a Promise-returning operation with required arguments, the generated wrapper throws TypeError if insufficient number of arguments are provided. I believe the expected behavior is to return a Promise.reject that throws.

E.g.

```webidl
interface Foo {
  Promise<string> greet(string name);
};
```

Generates:

```js
Foo.prototype.greet = function greet(name) {
  ...
  if (arguments.length < 1) {
    throw new TypeError(
      "Failed to execute 'greet' on 'Foo': " + "1 argument required, but only " + arguments.length + " present."
    );  // [*]
  }

  ...
}
```

[*] should be:

```js
return Promise.reject(new TypeError("<error message>"));
```

I couldn't find exact wording in WebIDL spec about this. However, the return-rejected-promise behavior is what Web Platform Test's idlharness.js enforces.

domenic:
Closes #79.

This removes support for overloading promise operations and non-promise operations, since that would require a more ambitious approach and doesn't seem to appear on the web.

---

Thoughts welcome. The strategy of using `async` for operations and try/catch for attributes makes the output nice but the webidl2js code a bit icky. Threading the async bit through all the operation stuff adds a bit of complexity. And, we might want to abandon that if we _do_ want to support overloads of promise-returning operations vs. non-promise returning ones, since doing that would require an appropriately-scoped try/catch.

### DIFF
--- a/lib/constructs/attribute.js
+++ b/lib/constructs/attribute.js
@@ -23,6 +23,10 @@ class Attribute {
 
     const onInstance = utils.isOnInstance(this.idl, this.interface.idl);
 
+    const async = this.idl.idlType.generic === "Promise";
+    const promiseHandlingBefore = async ? `try {` : ``;
+    const promiseHandlingAfter = async ? `} catch (e) { return Promise.reject(e); }` : ``;
+
     let brandCheck = `
       if (!exports.is(esValue)) {
         throw new TypeError("Illegal invocation");
@@ -72,12 +76,18 @@ class Attribute {
     }
 
     addMethod(this.idl.name, [], `
+      ${promiseHandlingBefore}
       const esValue = this !== null && this !== undefined ? this : globalObject;
       ${brandCheck}
       ${getterBody}
+      ${promiseHandlingAfter}
     `, "get", { configurable });
 
     if (!this.idl.readonly) {
+      if (async) {
+        throw new Error(`Illegal promise-typed attribute "${this.idl.name}" in interface "${this.interface.idl.name}"`);
+      }
+
       let idlConversion;
       if (typeof this.idl.idlType.idlType === "string" && !this.idl.idlType.nullable &&
           this.ctx.enumerations.has(this.idl.idlType.idlType)) {
--- a/lib/constructs/operation.js
+++ b/lib/constructs/operation.js
@@ -27,6 +27,22 @@ class Operation {
     return firstOverloadOnInstance;
   }
 
+  isAsync() {
+    // As of the time of this writing, the current spec does not disallow such overloads, but the intention is to do so:
+    // https://github.com/heycam/webidl/pull/776
+    const firstAsync = this.idls[0].idlType.generic === "Promise";
+    for (const overload of this.idls.slice(1)) {
+      const isAsync = overload.idlType.generic === "Promise";
+      if (isAsync !== firstAsync) {
+        throw new Error(
+          `Overloading between Promise and non-Promise return types is not allowed: operation ` +
+          `"${this.name}" on ${this.interface.name}`
+        );
+      }
+    }
+    return firstAsync;
+  }
+
   fixUpArgsExtAttrs() {
     for (const idl of this.idls) {
       for (const arg of idl.arguments) {
@@ -50,6 +66,9 @@ class Operation {
     }
 
     const onInstance = this.isOnInstance();
+    const async = this.isAsync();
+    const promiseHandlingBefore = async ? `try {` : ``;
+    const promiseHandlingAfter = async ? `} catch (e) { return Promise.reject(e); }` : ``;
 
     const type = this.static ? "static operation" : "regular operation";
     const overloads = Overloads.getEffectiveOverloads(type, this.name, 0, this.interface);
@@ -99,6 +118,8 @@ class Operation {
       });
     }
     str += invocation;
+
+    str = promiseHandlingBefore + str + promiseHandlingAfter;
 
     if (this.static) {
       this.interface.addStaticMethod(this.name, argNames, str);
--- a/test/__snapshots__/test.js.snap
+++ b/test/__snapshots__/test.js.snap
@@ -446,6 +446,24 @@ exports.install = (globalObject, globalNames) => {
       }
     }
 
+    promiseOperation() {
+      try {
+        const esValue = this !== null && this !== undefined ? this : globalObject;
+        if (!exports.is(esValue)) {
+          throw new TypeError(\\"Illegal invocation\\");
+        }
+
+        CEReactions.preSteps(globalObject);
+        try {
+          return utils.tryWrapperForImpl(esValue[implSymbol].promiseOperation());
+        } finally {
+          CEReactions.postSteps(globalObject);
+        }
+      } catch (e) {
+        return Promise.reject(e);
+      }
+    }
+
     get attr() {
       const esValue = this !== null && this !== undefined ? this : globalObject;
 
@@ -479,10 +497,31 @@ exports.install = (globalObject, globalNames) => {
         CEReactions.postSteps(globalObject);
       }
     }
+
+    get promiseAttribute() {
+      try {
+        const esValue = this !== null && this !== undefined ? this : globalObject;
+
+        if (!exports.is(esValue)) {
+          throw new TypeError(\\"Illegal invocation\\");
+        }
+
+        CEReactions.preSteps(globalObject);
+        try {
+          return utils.tryWrapperForImpl(esValue[implSymbol][\\"promiseAttribute\\"]);
+        } finally {
+          CEReactions.postSteps(globalObject);
+        }
+      } catch (e) {
+        return Promise.reject(e);
+      }
+    }
   }
   Object.defineProperties(CEReactions.prototype, {
     method: { enumerable: true },
+    promiseOperation: { enumerable: true },
     attr: { enumerable: true },
+    promiseAttribute: { enumerable: true },
     [Symbol.toStringTag]: { value: \\"CEReactions\\", configurable: true }
   });
   if (globalObject[ctorRegistrySymbol] === undefined) {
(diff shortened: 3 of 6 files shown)


domenic:
I found https://github.com/heycam/webidl/pull/776 which indicates we don't need to worry about supporting overloads between promise-returning and non-promise-returning operations.

lib/constructs/attribute.js
```diff
@@ -23,6 +23,10 @@ class Attribute {
 
     const onInstance = utils.isOnInstance(this.idl, this.interface.idl);
 
+    const async = this.idl.idlType.generic === "Promise";
+    const promiseHandlingBefore = async ? `try {` : ``;
+    const promiseHandlingAfter = async ? `} catch (e) { return Promise.reject(e); }` : ``;
```
ExE-Boss, line 28:
```suggestion
    const isAsync = this.idl.idlType.generic === "Promise";
    const promiseHandlingBefore = isAsync ? `try {` : ``;
    const promiseHandlingAfter = isAsync ? `} catch (e) { return Promise.reject(e); }` : ``;
```

lib/constructs/attribute.js
```diff
@@ -64,12 +68,18 @@ class Attribute {
...
+      ${promiseHandlingBefore}
       const esValue = this !== null && this !== undefined ? this : globalObject;
       ${brandCheck}
       ${getterBody}
-    `, "get", { configurable });
+      ${promiseHandlingAfter}
+    `, { type: "get" }, { configurable });
 
     if (!this.idl.readonly) {
+      if (async) {
```
ExE-Boss, line 79:
```suggestion
      if (isAsync) {
```

lib/constructs/interface.js
```diff
@@ -123,7 +123,7 @@ class Interface {
 
   // whence is either "instance" or "prototype"
   // type is either "regular", "get", or "set"
-  addMethod(whence, propName, args, body, type = "regular", {
+  addMethod(whence, propName, args, body, { type = "regular", async = false } = {}, {
```
ExE-Boss, line 126:
```suggestion
  addMethod(whence, propName, args, body, { type = "regular", isAsync = false } = {}, {
```

lib/constructs/interface.js
```diff
@@ -150,11 +150,11 @@ class Interface {
     }
 
     const descriptor = { configurable, enumerable, writable };
-    this._outputMethods.set(propName, { whence, type, args, body, descriptor });
+    this._outputMethods.set(propName, { whence, type, args, body, descriptor, async });
```
ExE-Boss, line 153:
```suggestion
    this._outputMethods.set(propName, { whence, type, args, body, descriptor, isAsync });
```

lib/constructs/interface.js
```diff
@@ -150,11 +150,11 @@ class Interface {
...
   }
 
   // type is either "regular", "get", or "set"
-  addStaticMethod(propName, args, body, type = "regular", {
+  addStaticMethod(propName, args, body, { type = "regular", async = false } = {}, {
```
ExE-Boss, line 157:
```suggestion
  addStaticMethod(propName, args, body, { type = "regular", isAsync = false } = {}, {
```

lib/constructs/interface.js
```diff
@@ -178,7 +178,7 @@ class Interface {
     }
 
     const descriptor = { configurable, enumerable, writable };
-    this._outputStaticMethods.set(propName, { type, args, body, descriptor });
+    this._outputStaticMethods.set(propName, { type, args, body, descriptor, async });
```
ExE-Boss, line 181:
```suggestion
    this._outputStaticMethods.set(propName, { type, args, body, descriptor, isAsync });
```

lib/constructs/interface.js
```diff
@@ -1311,40 +1311,40 @@ class Interface {
...
-    const addOne = (name, args, body) => {
+    const addOne = (name, args, body, { isAsync, isStatic }) => {
       this.str += `
-        ${name}(${formatArgs(args)}) {${body}}
+        ${isStatic ? "static " : ""}${isAsync ? "async " : ""}${name}(${formatArgs(args)}) {${body}}
       `;
     };
 
-    for (const [name, { whence, type, args, body }] of this._outputMethods) {
+    for (const [name, { whence, type, args, body, async }] of this._outputMethods) {
```
ExE-Boss, line 1320:
```suggestion
    for (const [name, { whence, type, args, body, isAsync }] of this._outputMethods) {
```

lib/constructs/interface.js
```diff
@@ -1311,40 +1311,40 @@ class Interface {
...
       if (whence !== "prototype") {
         continue;
       }
 
       const propName = utils.stringifyPropertyKey(name);
       if (type === "regular") {
-        addOne(propName, args, body);
+        addOne(propName, args, body, { isAsync: async });
```
ExE-Boss, line 1327:
```suggestion
        addOne(propName, args, body, { isAsync });
```

lib/constructs/interface.js
```diff
@@ -1311,40 +1311,40 @@ class Interface {
...
       } else {
         if (body[0] !== undefined) {
-          addOne(`get ${propName}`, [], body[0]);
+          addOne(`get ${propName}`, [], body[0], { isAsync: async });
```
ExE-Boss, line 1330:
```suggestion
          addOne(`get ${propName}`, [], body[0], { isAsync });
```

lib/constructs/interface.js
```diff
@@ -1311,40 +1311,40 @@ class Interface {
...
         }
         if (body[1] !== undefined) {
-          addOne(`set ${propName}`, args, body[1]);
+          addOne(`set ${propName}`, args, body[1], { isAsync: async });
```
ExE-Boss, line 1333:
```suggestion
          addOne(`set ${propName}`, args, body[1], { isAsync });
```

lib/constructs/interface.js
```diff
@@ -1311,40 +1311,40 @@ class Interface {
...
         }
       }
     }
 
-    for (const [name, { type, args, body }] of this._outputStaticMethods) {
+    for (const [name, { type, args, body, async }] of this._outputStaticMethods) {
```
ExE-Boss, line 1338:
```suggestion
    for (const [name, { type, args, body, isAsync }] of this._outputStaticMethods) {
```

lib/constructs/interface.js
```diff
@@ -1311,40 +1311,40 @@ class Interface {
...
       const propName = utils.stringifyPropertyKey(name);
       if (type === "regular") {
-        addOne(`static ${propName}`, args, body);
+        addOne(`${propName}`, args, body, { isAsync: async, isStatic: true });
```
ExE-Boss, line 1341:
```suggestion
        addOne(`${propName}`, args, body, { isAsync, isStatic: true });
```

lib/constructs/interface.js
```diff
@@ -1311,40 +1311,40 @@ class Interface {
...
       } else {
         if (body[0] !== undefined) {
-          addOne(`static get ${propName}`, [], body[0]);
+          addOne(`get ${propName}`, [], body[0], { isAsync: async, isStatic: true });
```
ExE-Boss, line 1344:
```suggestion
          addOne(`get ${propName}`, [], body[0], { isAsync, isStatic: true });
```

lib/constructs/interface.js
```diff
@@ -1311,40 +1311,40 @@ class Interface {
...
         }
         if (body[1] !== undefined) {
-          addOne(`static set ${propName}`, args, body[0]);
+          addOne(`set ${propName}`, args, body[0], { isAsync: async, isStatic: true });
```
ExE-Boss, line 1347:
```suggestion
          addOne(`set ${propName}`, args, body[0], { isAsync, isStatic: true });
```

lib/constructs/interface.js
```diff
@@ -1311,40 +1311,40 @@ class Interface {
   }
 
   generateOffInstanceMethods() {
-    const addOne = (name, args, body) => {
+    const addOne = (name, args, body, { isAsync, isStatic }) => {
```
ExE-Boss, line 1314:
```suggestion
    const addOne = (name, args, body, { isAsync = false, isStatic = false }) => {
```

lib/constructs/operation.js
```diff
@@ -48,6 +62,7 @@ class Operation {
     }
 
     const onInstance = this.isOnInstance();
+    const async = this.isAsync();
```
ExE-Boss, line 65:
```suggestion
    const isAsync = this.isAsync();
```

lib/constructs/operation.js
```diff
@@ -99,15 +114,15 @@ class Operation {
     str += invocation;
 
     if (this.static) {
-      this.interface.addStaticMethod(this.name, argNames, str);
+      this.interface.addStaticMethod(this.name, argNames, str, { async });
```
ExE-Boss, line 117:
```suggestion
      this.interface.addStaticMethod(this.name, argNames, str, { isAsync });
```

lib/constructs/operation.js
```diff
@@ -99,15 +114,15 @@ class Operation {
...
     } else {
       const forgeable = !utils.getExtAttr(this.idls[0].extAttrs, "Unforgeable");
       this.interface.addMethod(
         onInstance ? "instance" : "prototype",
         this.name,
         argNames,
         str,
-        "regular",
+        { type: "regular", async },
```
ExE-Boss, line 125:
```suggestion
        { type: "regular", isAsync },
```

test/__snapshots__/test.js.snap
```diff
@@ -2983,12 +3023,55 @@ exports.install = (globalObject, globalNames) => {
...
       return esValue[implSymbol].promiseConsumer(...args);
     }
+
+    async promiseOperation() {
+      const esValue = this !== null && this !== undefined ? this : globalObject;
+      if (!exports.is(esValue)) {
+        throw new TypeError(\\"Illegal invocation\\");
+      }
+
+      return utils.tryWrapperForImpl(esValue[implSymbol].promiseOperation());
```
ExE-Boss, line 3033: This has to generate:
```js
       return utils.tryWrapperForImpl(await esValue[implSymbol].promiseOperation());
```

test/__snapshots__/test.js.snap
```diff
@@ -2983,12 +3023,55 @@ exports.install = (globalObject, globalNames) => {
...
+
+    get promiseAttribute() {
+      try {
+        const esValue = this !== null && this !== undefined ? this : globalObject;
+
+        if (!exports.is(esValue)) {
+          throw new TypeError(\\"Illegal invocation\\");
+        }
+
+        return utils.tryWrapperForImpl(esValue[implSymbol][\\"promiseAttribute\\"]);
```
ExE-Boss, line 3044: This has to generate:
```js
          return Promise.resolve(esValue[implSymbol]["promiseAttribute"])
            .then(utils.tryWrapperForImpl);
```

test/__snapshots__/test.js.snap
```diff
@@ -2983,12 +3023,55 @@ exports.install = (globalObject, globalNames) => {
...
+      } catch (e) {
+        return Promise.reject(e);
+      }
+    }
+
+    static async staticPromiseOperation() {
+      return utils.tryWrapperForImpl(Impl.implementation.staticPromiseOperation());
```
ExE-Boss, line 3051: This has to generate:
```js
       return utils.tryWrapperForImpl(await Impl.implementation.staticPromiseOperation());
```

test/__snapshots__/test.js.snap
```diff
@@ -2983,12 +3023,55 @@ exports.install = (globalObject, globalNames) => {
...
+    }
+
+    static get staticPromiseAttribute() {
+      try {
+        const esValue = this !== null && this !== undefined ? this : globalObject;
+
+        return Impl.implementation[\\"staticPromiseAttribute\\"];
```
ExE-Boss, line 3058: This has to generate:
```js
          return Promise.resolve(Impl.implementation["staticPromiseAttribute"])
            .then(utils.tryWrapperForImpl);
```

domenic:
Please do not review pull requests that your review was not requested on. It is very rude to come into someone else's project and critique their code unsolicited.

test/__snapshots__/test.js.snap
domenic: This and your other unsolicited comments are not so straightforward.

You are suggesting that we incorporate the unwrapping done in https://github.com/jsdom/webidl2js/pull/108 here. However, that breaks the ability for the impl method/getter to return the same promise multiple times. That is especially problematic for attributes, where `x.attr === x.attr` is almost always what the spec intends.

I think we should leave working on the unwrapping to https://github.com/jsdom/webidl2js/pull/108 or some continuation of that has better discussion behind it.

lib/constructs/operation.js
```diff
@@ -25,6 +25,20 @@ class Operation {
     return firstOverloadOnInstance;
   }
 
+  isAsync() {
+    const firstAsync = this.idls[0].idlType.generic === "Promise";
```
TimothyGu, line 29: Linking to https://github.com/heycam/webidl/pull/776 would be useful.

lib/constructs/interface.js
```diff
@@ -1311,40 +1311,40 @@ class Interface {
...
       this.str += `
-        ${name}(${formatArgs(args)}) {${body}}
+        ${isStatic ? "static " : ""}${isAsync ? "async " : ""}${name}(${formatArgs(args)}) {${body}}
```
TimothyGu, line 1316: Do we actually need the function to be async? I don't see `await` being used anywhere, and I fear that this difference could in fact be observable by the number of ticks it passed before the promise is rejected with a synchronously thrown error.

ExE-Boss: I talk about that in <https://github.com/jsdom/webidl2js/issues/79#issuecomment-617738500>.

domenic: It's a convenient way to convert exceptions into rejections. But, it sounds like folks are in favor of instead wrapping everything in try/catch, so I'll do that.

test/cases/PromiseTypes.webidl
```diff
@@ -2,4 +2,13 @@
 interface PromiseTypes {
   void voidPromiseConsumer(Promise<void> p);
   void promiseConsumer(Promise<double> p);
+
+  Promise<void> promiseOperation();
+  readonly attribute Promise<void> promiseAttribute;
+
+  [Unforgeable] Promise<void> unforgeablePromiseOperation();
+  [Unforgeable] readonly attribute Promise<void> unforgeablePromiseAttribute;
```
TimothyGu, line 10:
```suggestion
  [LegacyUnforgeable] Promise<void> unforgeablePromiseOperation();
  [LegacyUnforgeable] readonly attribute Promise<void> unforgeablePromiseAttribute;
```

PR #201, opened 2020-04-18
Attempt to fix the v15 release line to not be backward-compat breaking

domenic:
This attempts to revert all changes in master plus v15.2.0 that were backward-compat breaking, so that we can release a v15.2.1 that backward-compatible with v15.1.0 and v15.0.0. Such a v15.2.1 would have the following deltas from v15.1.0:

- Full support for callback interfaces
- Updated webidl-conversions dependency

(The addition of `new()` is reverted for simplicity, since it was based on the removal of `init()`.)

It would be nice if we could unpublish v15.2.0, but apparently we can't since it has dependent packages in the registry.

The intention would then be to un-revert these four commits (except maybe not the removal of `init()`; see https://github.com/jsdom/webidl2js/pull/195#pullrequestreview-395813297) and publish a v16.0.0 with the breaking changes of requiring `[Exposed]` on all interfaces and requiring a second argument to `install()`.

This is the best I can currently think of, but I welcome a second pair of eyes, since I already messed this up.


lib/constructs/callback-interface.js
```diff
@@ -197,24 +178,14 @@ class CallbackInterface {
...
   generateInstall() {
-    if (this.constants.size > 0) {
-      this.str += `
-        const exposed = new Set(${JSON.stringify([...this.exposed])});
-      `;
-    }
-
     this.str += `
-      exports.install = (globalObject, globalNames) => {
+      exports.install = function install(globalObject) {
```
ExE-Boss, line 182:
```suggestion
      exports.install = (globalObject, globalNames = ["Window"]) => {
```

test/cases/CEReactions.webidl
```diff
@@ -1,8 +1,7 @@
-[Exposed=Window]
 interface CEReactions {
     [CEReactions] attribute DOMString attr;
     [CEReactions] void method();
-
+    
     getter DOMString (DOMString name);
```
ExE-Boss, line 5:
```suggestion

    getter DOMString (DOMString name);
```

test/cases/CEReactions.webidl
```diff
@@ -1,8 +1,7 @@
-[Exposed=Window]
 interface CEReactions {
```
ExE-Boss, line 1:
```suggestion
[Exposed=Window]
interface CEReactions {
```

test/cases/Global.webidl
```diff
@@ -1,4 +1,4 @@
-[Global=Global,Exposed=Global]
+[Global=Global,Exposed=Window]
```
ExE-Boss, line 1:
```suggestion
[Global=Global,Exposed=Global]
```

test/cases/HTMLConstructor.webidl
```diff
@@ -1,3 +1,2 @@
-[Exposed=Window,
- HTMLConstructor]
+[HTMLConstructor]
```
ExE-Boss, line 1:
```suggestion
[Exposed=Window,
 HTMLConstructor]
```

test/cases/URL.webidl
```diff
@@ -1,5 +1,4 @@
-[Exposed=(Window,Worker),
- LegacyWindowAlias=webkitURL]
+[Exposed=(Window,Worker)]
```
ExE-Boss, line 1:
```suggestion
[Exposed=(Window,Worker),
 LegacyWindowAlias=webkitURL]
```

ExE-Boss:
Alternatively, just create a new branch `v15` and backport things to it.

domenic:
Hey @ExE-Boss, please refrain from reviewing in the future; your review was not requested and it's not polite to do drive-by reviews in this way.

Making globalNames optional will not solve the backward-compat issue that [Exposed] is now required, I believe.

domenic:
Replaced by https://github.com/jsdom/webidl2js/pull/202.

PR #206, opened 2020-04-20
Align with Web IDL extended attribute renames

domenic:
I added a second commit for the renaming, so that you can see that the first commit does not effect the snapshot output at all.

Interestingly, we seem to support [LegacyLenientThis] and [LegacyOverrideBuiltins] with no tests.

This conflicts with #203; I will take care of the conflicts when merging, if both get approved.

### DIFF
--- a/README.md
+++ b/README.md
@@ -469,15 +469,16 @@ webidl2js is implementing an ever-growing subset of the Web IDL specification. S
 - `[Clamp]`
 - `[EnforceRange]`
 - `[Exposed]`
+- `[LegacyLenientThis]`
+- `[LegacyNoInterfaceObject]`
+- `[LegacyNullToEmptyString]`
+- `[LegacyOverrideBuiltins]`
 - `[LegacyUnenumerableNamedProperties]`
+- `[LegacyUnforgeable]`
 - `[LegacyWindowAlias]`
-- `[NoInterfaceObject]`
-- `[OverrideBuiltins]`
 - `[PutForwards]`
 - `[Replaceable]`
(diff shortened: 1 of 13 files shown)


lib/utils.js
```diff
@@ -38,7 +38,7 @@ function hasCEReactions(idl) {
 }
 
 function isOnInstance(memberIDL, interfaceIDL) {
-  return memberIDL.special !== "static" && (getExtAttr(memberIDL.extAttrs, "Unforgeable") || isGlobal(interfaceIDL));
+  return memberIDL.special !== "static" && (getExtAttr(memberIDL.extAttrs, "LegacyUnforgeable") || isGlobal(interfaceIDL));
```
ExE-Boss, line 41: This line is too long, which is why the build is failing:

```suggestion
  return (
    memberIDL.special !== "static" &&
    (getExtAttr(memberIDL.extAttrs, "LegacyUnforgeable") || isGlobal(interfaceIDL))
  );
```

The above is output from **Prettier**.

README.md
```diff
@@ -492,11 +493,11 @@ Notable missing features include:
 - `[AllowShared]`
 - `[Default]` (for `toJSON()` operations)
 - `[Global]`'s various consequences, including the named properties object and `[[SetPrototypeOf]]`
-- `[LenientSetter]`
-- `[LenientThis]`
```
TimothyGu, line 496: I seem to remember `LenientThis` was not moved to the list of implemented features because it was implemented in some places but not others. I can't find anything now though so I suppose it's since been implemented.

Issue #188, opened 2020-04-01
Support adding dynamic IDL strings to Transformer


ExE-Boss:
This is a pain point I’ve encountered while trying to convert `cssstyle` to use **WebIDL2JS**.

I’m currently forced to generate `./src/CSSStyleDeclaration‑properties.webidl` at build time, which contains the generated `partial interface` attributes:
- https://drafts.csswg.org/cssom/#ref-for-cssstyledeclaration%E2%91%A3
- https://drafts.csswg.org/cssom/#ref-for-cssstyledeclaration%E2%91%A4
- https://drafts.csswg.org/cssom/#ref-for-cssstyledeclaration%E2%91%A5

TimothyGu:
For reference, Blink just uses a getter/setter rather than implementing each attribute individually: https://source.chromium.org/chromium/chromium/src/+/master:third_party/blink/renderer/core/css/css_style_declaration.idl;l=1;drc=fdd99242518b07d1d48e6f8453cbb7963a27cb6a.

Gecko also has an IDL generation step: see https://searchfox.org/mozilla-central/source/dom/bindings/GenerateCSS2PropertiesWebIDL.py.

domenic:
Generating IDL at build time, like Gecko, seems like the right approach to me.

ExE-Boss:
I also generate them at build time, but I’d like to skip the writing to disk part.

domenic:
That makes some sense, and perhaps should feed in to #160.

Issue #160, opened 2020-01-25
Consider a better API for feeding in files


domenic:
https://github.com/jsdom/webidl2js/pull/159 ran into an issue where the current API is not flexible enough to do targeted testing, because it operates on a directory level. This has always been a bit of a strange design; it's not very flexible and assumes a lot about the directory structure. I'd love to get opinions on what would be better designs.

Here are some ideas:

```js
const transformer = new WebIDL2JS();

transformer.addSource("path/to/Attr.webidl", "path/to/Attr-impl.js");
// ... add the rest, one file at a time. The client is responsible for directory iteration.

await transformer.generate("outputDir");
```

```js
const transformer = new WebIDL2JS({
  implPath(webidlFilePath) { return ...; }
  wrapperPath(webidlFilePath, implPath) { return ...; }
});

transformer.addFile("webidlFilesDir/Attr.webidl");

// Optionally add some of these as convenience methods:
transformer.addDirectory("webidlFilesDir");
transformer.addFromGlob("webidlFilesDir/**.webidl");

await transformer.generate();
```

Other ideas are welcome as well.

ExE-Boss:
In <https://github.com/jsdom/webidl2js/pull/208>, I implemented `addFile(…)` and `addDirectory(…)` to match `addSource(…)`, except with an optional `impl` path argument.

---

It also fixes the bug where:

```js
const transformer = new WebIDL2JS();

transformer.addSource("path/to/Attr.webidl", "path/to/Attr-impl.js");
// ... add the rest, one file at a time. The client is responsible for directory iteration.

await transformer.generate("outputDir");
```

Would result in `outputDir/Attr.js` requiring `path/to/Attr‑impl.js/Attr.js` instead of `path/to/Attr‑impl.js`, because the wrapper generation code currently assumes that the `impl` path argument is always a directory.

domenic:
I'd like us to consider and decide on the eventual shape of the whole API, and not just add things. In particular, I think `addSource` is a bad API with a bad name.

Can you describe how `addFile` and `addDirectory` work, including what arguments they take? And, if you disagree and think `addSource` should be kept, then can you include that as part of your description? Basically, before we consider working on a pull request, I'd like to see a concrete proposal, at least at the level of detail of API documentation.

domenic:
Testing would be easier if we had a core API that didn't do file I/O at all, plus a wrapper API that did the file I/O. I guess that would be something like

```js
const transformer = new WebIDL2JS();

[...]

PR #208, opened 2020-04-22
Implement `addFile(…)` and `addDirectory(…)` Transformer APIs

ExE-Boss:
Part of #160

---

I implemented `addFile(…)` and `addDirectory(…)` to match `addSource(…)`, except with an optional `impl` path argument.

This also fixes the bug where:
```js
const transformer = new WebIDL2JS();

transformer.addSource("path/to/Attr.webidl", "path/to/Attr-impl.js");
// ... add the rest, one file at a time. The client is responsible for directory iteration.

await transformer.generate("outputDir");
```

Would result in `outputDir/Attr.js` requiring `path/to/Attr‑impl.js/Attr.js` instead of `path/to/Attr‑impl.js`, because the wrapper generation code currently assumes that the `impl` path argument is always a directory.

This doesn’t yet implement `addFromGlob` or a way to feed in dynamically generated `partial interface` definitions (<https://github.com/jsdom/webidl2js/issues/188>).

---

This will also make it possible to implement <https://github.com/jsdom/jsdom/pull/2835> without renaming `Window.js` (https://github.com/jsdom/jsdom/pull/2837) by utilising the above mention `impl` path name fix.

### DIFF
--- a/lib/context.js
+++ b/lib/context.js
@@ -21,7 +21,7 @@ class Context {
     processHTMLConstructor = defaultProcessor,
     processReflect = null,
     options
-  } = {}) {
+  }) {
     this.implSuffix = implSuffix;
     this.processCEReactions = processCEReactions;
     this.processHTMLConstructor = processHTMLConstructor;
--- a/lib/transformer.js
+++ b/lib/transformer.js
@@ -26,7 +26,7 @@ class Transformer {
       }
     });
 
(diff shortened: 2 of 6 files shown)


domenic:
Can you discuss your proposal in #160 first?

domenic:
Let's close this until we have an agreed-upon design for the API in #160.

Issue #207, opened 2020-04-20
[LegacyLenientThis] and [LegacyOverrideBuiltins] are supported but don't have tests


domenic:
We should fix this, after #206 lands.

ExE-Boss:
Same with `[Replaceable]`.

Also, `[LegacyLenientThis]` currently has a bug where the getter (and setter) will be invoked regardless of whether or not `esValue` implements the interface, I’m fixing that in <https://github.com/jsdom/webidl2js/pull/209>.

domenic:
I've renamed this issue and updated the OP to cover all untested extended attributes.

PR #209, opened 2020-04-22
Implement `[LegacyLenientSetter]` and test `[LegacyLenientThis]`

ExE-Boss:
This also fixes the bug where `[LegacyLenientThis]` would still result in attempting to get the implementation value, even when `exports.is(esValue)` was `false`, which would result in a `TypeError` if `esValue` did not have an `implSymbol` property.

---

This will have merge conflicts with <https://github.com/jsdom/webidl2js/pull/206> and should probably be rebased on top of it, since it doesn’t make sense to have `[LegacyLenientThis]` use the new naming scheme, but have `[LegacyLenientSetter]` using the old naming scheme.

---

Part of <https://github.com/jsdom/webidl2js/issues/207>

### DIFF
--- a/lib/constructs/attribute.js
+++ b/lib/constructs/attribute.js
@@ -48,8 +48,16 @@ class Attribute {
       setterBody = processedOutput.set;
     }
 
-    if (utils.getExtAttr(this.idl.extAttrs, "LenientThis")) {
-      brandCheck = "";
+    const replaceable = utils.getExtAttr(this.idl.extAttrs, "Replaceable");
+    const legacyLenientSetter = utils.getExtAttr(this.idl.extAttrs, "LegacyLenientSetter");
+    const legacyLenientThis = utils.getExtAttr(this.idl.extAttrs, "LenientThis");
+
+    if (legacyLenientThis) {
+      brandCheck = `
+        if (!exports.is(esValue)) {
+          return;
+        }
+      `;
     }
 
     if (sameObject) {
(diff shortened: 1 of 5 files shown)


test/cases/LegacyLenientAttributes.webidl
```diff
@@ -0,0 +1,7 @@
+[Exposed=Window]
+interface LegacyLenientAttributes {
+  [LegacyLenientSetter] readonly attribute DOMString lenientSetter;
+  [LegacyLenientSetter, LenientThis] readonly attribute DOMString lenientThisSetter;
+  [LenientThis] readonly attribute DOMString lenientThis;
+  [LenientThis, Replaceable] readonly attribute DOMString replaceableLenientThis;
```
domenic, line 6: Can you also test

- `[LenientThis]` by itself, non-readonly attribute
- `[Replaceable]` by itself, readonly attribute

?

lib/constructs/attribute.js
```diff
@@ -48,8 +48,16 @@ class Attribute {
       setterBody = processedOutput.set;
     }
 
-    if (utils.getExtAttr(this.idl.extAttrs, "LenientThis")) {
-      brandCheck = "";
+    const replaceable = utils.getExtAttr(this.idl.extAttrs, "Replaceable");
+    const legacyLenientSetter = utils.getExtAttr(this.idl.extAttrs, "LegacyLenientSetter");
+    const legacyLenientThis = utils.getExtAttr(this.idl.extAttrs, "LenientThis");
```
domenic, line 53: As a bonus, can you enforce (via compile-time errors) the following:

> An attribute with the [Replaceable] extended attribute must not also be declared with the [LegacyLenientSetter] or [PutForwards] extended attributes.

> The [LegacyLenientSetter] extended attribute must take no arguments. It must not be used on anything other than a read only regular attribute.

> An attribute with the [LegacyLenientSetter] extended attribute must not also be declared with the [PutForwards] or [Replaceable] extended attributes.

> The [LegacyLenientThis] extended attribute must take no arguments. It must not be used on a static attribute.

> The [Replaceable] extended attribute must take no arguments.

> An attribute with the [Replaceable] extended attribute must not also be declared with the [LegacyLenientSetter] or [PutForwards] extended attributes.

> The [Replaceable] extended attribute must not be used on an attribute that is not read only.

> The [Replaceable] extended attribute must not be used on a static attribute.

(I've omitted the not-on-a-namespace checks since we don't do namespaces.)

Enforcing these would make it more obvious why we don't have tests for some of these cases.

If you would rather do these in a separate PR, perhaps including checks for all the other extended attributes we implement, that is fine too. Perhaps at that time we could even introduce a way of testing these errors.

ExE-Boss: I intend to do that in its own PR.

domenic:
Closing since this is from an organization and we don't accept PRs from organizations anymore. I'll reopen a new PR instead.

Issue #186, opened 2020-03-27
Pass global object to static methods


TimothyGu:
This is needed for things like `DOMRect`'s [`fromRect()`](https://drafts.fxtf.org/geometry/#dom-domrect-fromrect).

domenic:
Continuing the conversation from #211: I think I like Chromium's mechanism, where by default you don't get the global object passed, but you can opt in with an extended attribute. I'm thinking `[WebIDL2JSCallWithGlobal]` which makes the first arugment always be the global object.

What do you think?

TimothyGu:
That works.

Issue #199, opened 2020-04-17
Static setter code is broken


domenic:
It generates the same body as the getter, it seems:

```
    static set abc(V) {
      const esValue = this !== null && this !== undefined ? this : globalObject;

      return Impl.implementation[\\"abc\\"];
    }
```

PR #211, opened 2020-04-28
Fix static attribute setters

ExE-Boss:
Fixes <https://github.com/jsdom/webidl2js/issues/199>

### DIFF
--- a/lib/constructs/interface.js
+++ b/lib/constructs/interface.js
@@ -1350,7 +1350,7 @@ class Interface {
           addOne(`static get ${propName}`, [], body[0]);
         }
         if (body[1] !== undefined) {
-          addOne(`static set ${propName}`, args, body[0]);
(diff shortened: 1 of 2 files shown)


TimothyGu:
This looks good, but I'm wondering if we can fix https://github.com/jsdom/webidl2js/issues/186 at the same time. I'm not entirely sure what the shape of the API should look like though… maybe `Impl.implementation.set${camelCase(attr)}(globalObject, value)`?

domenic:
Chromium does it via `[CallWith=]` nonstandard extended attributes, so that the impl class opts in to the unusual parameters.

domenic:
Anyway, my vote would be for merging this without blocking on the additional scope of #186. WDYT, @TimothyGu?

TimothyGu:
I was hoping to get something decided before a major version bump, but that ship has sailed so this is fine for me.

PR #212, opened 2020-04-29
Fix `addSource` to allow implementation path to point to a file

ExE-Boss:
This fixes the bug where:

```js
const transformer = new WebIDL2JS();

transformer.addSource("path/to/Attr.webidl", "path/to/Attr-impl.js");
// ... add the rest, one file at a time. The client is responsible for directory iteration.

await transformer.generate("outputDir");
```

Would result in `outputDir/Attr.js` requiring `path/to/Attr‑impl.js/Attr.js` instead of `path/to/Attr‑impl.js`, because the wrapper generation code currently assumes that the `impl` path argument is always a directory.

---

This will also make it possible to implement <https://github.com/jsdom/jsdom/pull/2835> without renaming `Window.js` (<https://github.com/jsdom/jsdom/pull/2837>) by utilising the above mention `impl` path name fix.

### DIFF
--- a/lib/transformer.js
+++ b/lib/transformer.js
@@ -26,7 +26,7 @@ class Transformer {
       }
     });
 
-    this.sources = []; // Absolute paths to the IDL and Impl directories.
+    this._sources = []; // Absolute paths to the IDL and Impl files or directories.
     this.utilPath = null;
   }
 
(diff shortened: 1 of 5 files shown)


domenic:
As I expressed in #208 and #160, I'd like us to decide on an API through discussion, before starting any pull requests.

PR #213, opened 2020-04-29
Implement [LegacyFactoryFunction]

ExE-Boss:
This implements [`[LegacyFactoryFunction]`](https://heycam.github.io/webidl/#LegacyFactoryFunction).

Unlike Chromium’s implementation, which assumes that a single interface will only ever have at most one legacy factory function, this implementation makes it possible to define more than one legacy factory function per interface, as can be seen with `test/cases/LegacyFactoryFunction.webidl`.


domenic:
> Unlike Chromium’s implementation, which assumes that a single interface will only ever have at most one legacy factory function, this implementation makes it possible to define more than one legacy factory function per interface, as can be seen with test/cases/LegacyFactoryFunction.webidl.

Let's remove that complexity. We can add a constraint to the Web IDL spec that prohibits more than one LegacyFactoryFunction instead.

domenic:
The generated snapshots here seem overcomplicated, and it's not clear why you're modifying existing functionality (like `new()`).

I think the root of the problem is that you're trying to construct the `thisArg`. Instead, since it's a factory function, the impl class can do that. In other words, I'd expect the generated code

```js
    const thisArgument = exports.new(globalObject, new.target);
    const result = Impl.legacyFactoryFunction.call(thisArgument, globalObject, ...args);
    return utils.tryWrapperForImpl(utils.isObject(result) ? result : thisArgument);
```

to instead be

```js
    return utils.tryWrapperForImpl(Impl.legacyFactoryFunction(globalObject, ...args));
```

ExE-Boss:
@domenic That won’t support the case when `new.target !== activeFunctionObject`.

domenic:
Hmm. I see there are references to NewTarget in https://heycam.github.io/webidl/#legacy-factory-functions but I am unsure whether they are implemented, or if they are, whether they're necessary for web compatibility.

ExE-Boss:
@domenic All browsers implement the **NewTarget** handling: <https://github.com/web-platform-tests/wpt/pull/27991>

So it’s probably necessary for web compatibility.

domenic:
@ExE-Boss want to rebase this and finish review and then we can do a new release?

README.md
```diff
@@ -421,6 +421,17 @@ It is often useful for implementation classes to inherit from each other, if the
 
 However, it is not required! The wrapper classes will have a correct inheritance chain, regardless of the implementation class inheritance chain. Just make sure that, either via inheritance or manual implementation, you implement all of the expected operations and attributes.
 
+### The `[LegacyFactoryFunction]` extended attribute
+
+For interfaces which have the `[LegacyFactoryFunction]` extended attribute, the implementation class file must contain the `legacyFactoryFunction` export, with the signature `(thisValue, globalObject, [...legacyFactoryFunctionArgs], { newTarget, wrapper })`, which is used for:
```
ExE-Boss, line 426: I decided to pass the `thisValue` as `arguments[0]`, rather than `this` so that it’s possible to use arrow functions for this:
```js
exports.legacyFactoryFunction = (thisValue, globalObject, [arg0, arg1], { newTarget }) => {
	// code
}
```

I also decided to include the `new.target` and `wrapper` in a `privateData` object to match `constructor(globalObject, args, privateData)`.

test/cases/OverloadedLegacyFactoryFunctionInterface.webidl
```diff
@@ -0,0 +1,4 @@
+[Exposed=Window,
+ LegacyFactoryFunction=OverloadedLegacyFactoryFunction(optional DOMString src),
+ LegacyFactoryFunction=OverloadedLegacyFactoryFunction(long arg1, long arg2)]
```
domenic, line 3: Let's disallow any instance of the attribute appearing more than once, per https://github.com/heycam/webidl/issues/878#issuecomment-803685451.

test/cases/HTMLOptionElement.webidl
```diff
@@ -0,0 +1,6 @@
+// Simplified from https://html.spec.whatwg.org/multipage/form-elements.html#htmloptionelement
```
domenic, line 1: What value does having these three tests give, instead of just one of them? I can see zero-args vs. nonzero, but these three cases don't seem to differ in interesting ways.

README.md
domenic: Would jsdom's implementations actually any of thisValue, newTarget, or wrapper? If not, let's exclude them.

lib/constructs/interface.js
```diff
@@ -1644,6 +1667,73 @@ class Interface {
...
+      function ${name}(${utils.formatArgs(args)}) {
+        if (new.target === undefined) {
+          throw new TypeError("Class constructor ${name} cannot be invoked without 'new'");
+        }
+
+        ${conversions.body}
+    `;
+
+    // This implements the WebIDL legacy factory function behavior, as well as support for overridding
+    // the return type, which is used by HTML's element legacy factory functions:
```
domenic, line 1712: This seems more likely to be a historical artifact of how the spec is written?

It seems like we should be able to support either return overload (like HTML uses for its LegacyFactoryFunctions), or generate a this value (like modern Web IDL-using constructor specs do). We shouldn't need both, I don't think?

In particular given that you can subclass `Option` per the tests you wrote, I suspect the first line of https://html.spec.whatwg.org/#dom-option is just incorrect.

domenic (changes requested):

lib/constructs/interface.js
ExE-Boss: It occurred to me that the order of `Impl.legacyFactoryFunction` and `Impl.init` calls is wrong, so I’m fixing that so that the object returned from `Impl.legacyFactoryFunction` is now the `impl`, and the result of `makeWrapper` is what’s returned from the `LegacyFactoryFunction` constructor.

README.md
ExE-Boss: We should at the very least include `thisValue` and pass the `args` as an array to leave open the possibility of a `privateData` bag in case <https://html.spec.whatwg.org/multipage/media.html#dom-audio>, <https://html.spec.whatwg.org/multipage/embedded-content.html#dom-image> and <https://html.spec.whatwg.org/multipage/form-elements.html#dom-option> being rewritten.

domenic: I'm still feeling like this is over-complex given that there are literally three cases and we know exactly how they behave. Our goal here is not to have a general framework to do any possible LegacyFactoryFunction; it's to implement Option, Audio, and Image.

In particular, I'd like to see these either support return override, or use the passed in thisValue, but not both. I suspect thisValue would be simpler, and perhaps necessary to support `class X extends Audio {}`. So I suspect the correct signature here will be `(thisValue, globalObject, [...args])` with no support for return-override.

PR #218, opened 2020-05-01
Fix legacy platform object [[Set]] operation

ExE-Boss:
This fixes the legacy platform object `[[Set]]` operation implementation to match the **WebIDL** spec, which doesn’t exclude array index properties from named properties when indexed setters aren't supported.

---

It also fixes the bug that was caused by the `receiver` argument in a Proxy trap pointing to the Proxy exotic object or an object that inherits from it, rather than to the underlying Proxy target.

This can be verified with:
```js
(new Proxy({}, { get(target,_,receiver) { console.log(target === receiver); } })).foo;
```
Which logs `false`.

---

review?(@domenic)

### DIFF
--- a/lib/constructs/interface.js
+++ b/lib/constructs/interface.js
@@ -841,7 +841,9 @@ class Interface {
           if (typeof P === "symbol") {
             return Reflect.set(target, P, V, receiver);
           }
-          if (target === receiver) {
+          // The \`receiver\` argument refers to the Proxy exotic object or an object
+          // that inherits from it, whereas \`target\` refers to the Proxy target:
+          if (target[implSymbol][utils.wrapperSymbol] === receiver) {
     `;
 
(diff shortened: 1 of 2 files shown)


domenic:
What does this impact? Can we add a web platform test for it?

ExE-Boss:
It causes the `webstorage/set.window.js` test to fail.

ExE-Boss:
@TimothyGu The `#` symbol refers to an in issue, to refer to a commit, you have to use the <https://github.com/heycam/webidl/commit/059491b4de91fdc45d26d9f2c48a6a6a0c3eb3e6> syntax.

Also, array index properties weren’t excluded before <https://github.com/heycam/webidl/commit/059491b4de91fdc45d26d9f2c48a6a6a0c3eb3e6> either.

PR #219, opened 2020-05-02
Fix ECMAScript to IDL `Promise` type conversion

ExE-Boss:
Most notably, this fixes the bug where rejected promises were being converted into resolved promises and that `Promise.resolve(x)` returns `x` when `x` is a `Promise` object from the same realm as the `Promise.resolve` function, whereas **WebIDL** requires the conversion to always return a new `Promise` object.

---

Note that there still is a small timing difference for non‑`void` promises due to the fact that the IDL conversion and promise reaction are being performed in separate `then` callbacks, whereas `WebIDL` runs them in the same `then` callback. (See the [“to **react** to a <code>Promise&lt;<var>T</var>&gt;</code>”](https://heycam.github.io/webidl/#dfn-perform-steps-once-promise-is-settled) algorithm)

Fixing this minor timing discrepancy would require the type conversion to be performed manually in implementation classes or overriding the `then` method of the converted `Promise` instance.

---

**P.S.:** If you still want `then` to be called with two arguments, then the correct callback is:
```js
reason => {
	throw reason;
}
```

Which is the default behaviour when `null` or `undefined` is passed.

### DIFF
--- a/lib/types.js
+++ b/lib/types.js
@@ -328,19 +328,39 @@ function generateTypeConversion(ctx, name, idlType, argAttrs = [], parentName, e
   }
 
   function generatePromise() {
-    str += `${name} = new Promise(resolve => resolve(${name}))`;
+    str += `${name} = new Promise(resolve => resolve(${name}));`;
 
     if (idlType.idlType[0].idlType !== "void") {
-      const conv = generateTypeConversion(ctx, "value", idlType.idlType[0], [], parentName,
+      const convOnFulfilled = generateTypeConversion(ctx, "value", idlType.idlType[0], [], parentName,
         `${errPrefix} + " promise value"`);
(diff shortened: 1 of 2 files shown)


domenic:
Can you remove the T conversion entirely? As I pointed out, it does not exist in the Web IDL spec.

ExE-Boss:
It does exist in the [“to react to a <code>Promise&lt;<var>T</var>&gt;</code>”](perform-steps-once-promise-is-settled) algorithm.

domenic:
Right. When the spec reacts to it, it can do the conversion. We shouldn't do it automatically.

domenic:
I'm sorry, perhaps I wasn't clear enough. The type conversion logic should not live in webidl2js at all.

PR #220, opened 2020-05-02
fix: setting of enumerated attribute case-sensitive

zjffun:
[keywords-and-enumerated-attributes](https://html.spec.whatwg.org/multipage/common-microsyntaxes.html#keywords-and-enumerated-attributes)

> If an enumerated attribute is specified, the attribute's value must be an ASCII case-insensitive match for one of the given keywords that are not said to be non-conforming, with no leading or trailing whitespace.


domenic:
This isn't correct. The section you're referring to is about enumerated attributes in HTML. It is not about enumerations in Web IDL.

zjffun:
My wrong. 😳

Issue #217, opened 2020-04-30
Better abstract string concatenation


TimothyGu:
Due to the way we construct type conversions, oftentimes we see unidiomatic code like https://github.com/jsdom/webidl2js/blob/7d45dfc6e6d1ba737b77eaabbab712dcf2d19435/test/__snapshots__/test.js.snap#L167 Ideally we want to merge string literals together and even use template literals.

Related to this is #204, but this seems mostly orthogonal and more tractable.

domenic:
I wonder if this could be done as a post-processing step using eslint --fix or some more aggressive prettier configuration.

TimothyGu:
Prettier has a stance of not changing ASTs, so I guess that'd be off. ESLint has [no-useless-concat](https://eslint.org/docs/rules/no-useless-concat), but sadly no fixer for it. This is easy to do even without a postprocessor though, so maybe we should just write the code.

ExE-Boss:
I already have code for this in <https://github.com/ExE-Boss/webidl2js/tree/refactor/precompute-concatenated-strings>, but I wasn’t sure whether it was something that was needed, and now it needs rebasing.

PR #221, opened 2020-05-03
Precompute concatenated strings where possible

ExE-Boss:
Fixes <https://github.com/jsdom/webidl2js/issues/217>


lib/constructs/iterable.js
```diff
@@ -39,12 +39,10 @@ class Iterable {
           throw new TypeError("Illegal invocation");
         }
         if (arguments.length < 1) {
-          throw new TypeError("Failed to execute 'forEach' on '${this.name}': 1 argument required, " +
-                              "but only 0 present.");
+          throw new TypeError("Failed to execute 'forEach' on '${this.name}': 1 argument required, but only 0 present.");
```
TimothyGu, line 42: I'd rather keep this the way it is.

Issue #204, opened 2020-04-20
Improve error messages for enums inside dictionaries


domenic:
Currently, converting `{ type: "asdf" }` to a dictionary defined as follows

```webidl
dictionary UnderlyingSource {
  ReadableStreamType type;
};

enum ReadableStreamType { "bytes" };
```

will produce an error message like

> The provided value has member 'type' that 'asdf' is not a valid enumeration value for ReadableStreamType

which is not grammatical. This arises because of how we compose our dictionary error message and enum error message.

Note that just trying to convert `"asdf"` to `ReadableStreamType` would produce something like

> The provided value 'asdf' is not a valid enumeration value for ReadableStreamType

Here are some thoughts:

- Is it a good idea to mention enum names in error messages at all? They're not web-developer-facing, so maybe they should just be omitted, both in dictionaries and outside.
- If we were to keep it, some ideas:
  - > 'asdf' is not a valid enumeration value for ReadableStreamType (provided as the 'type' member of the provided value)
  - > 'asdf' is not a valid ReadableStreamType enumeration value for the 'type' member of the provided value

PR #222, opened 2020-05-03
Improve “Illegal invocation” error messages

ExE-Boss:
The improved error message format is based on the one used in **Firefox**.

---

Related to <https://github.com/jsdom/webidl2js/issues/204> and <https://github.com/jsdom/webidl2js/issues/217>.


ExE-Boss:
I went with: “`'bar' called on an object that is not a valid instance of Foo.`”, because `Object.create(Foo.prototype)` is an instance of `Foo`, but the **WebIDL** brand checks will fail.

Issue #215, opened 2020-04-29
Support async iterables


domenic:
For https://github.com/whatwg/streams/pull/1035 it'd be great if webidl2js supported async iterators.

Note that our parser (webidl2.js) supports async value iterators but not yet arguments to async iterators: see the PRs linked in https://github.com/w3c/webidl2.js/issues/464.

Inspecting `Headers-impl.js` / the generated `Headers.js` in jsdom, our current iterator support works by having the impl class supply a `[Symbol.iterator]()` method. Then, in the generated `next()` method, it does `Array.from(impl)`, snapshotting the entire thing as an array. Finally it returns the appropriately-indexed value in the array. I don't think that strategy is great for async iterators. (Nor is it especially good for sync iterators...)

One idea is to follow the spec structure entirely. We'd add new symbols to `utils.js`:

- `utils.getTheNextIterationResult`
- `utils.asyncIteratorReturn`
- `utils.asyncIteratorInit`

and the generated code would call into these. This would allow fairly straightforward implementation of https://whatpr.org/streams/1035.html#rs-asynciterator .

There might be something more clever we could do with impl-class `[Symbol.asyncIterator]` methods, taking advantage of their coroutine nature. For example, maybe this would work?

```js
class ReadableStreamImpl {
  [Symbol.asyncIterator]({ preventCancel }) {
    // Async iterator initialization steps go here, before any yields or awaits. E.g.:
    const reader = AcquireReadableStreamDefaultReader(this);

    let returnWasCalled = true;
    try {
      while (true) {
        // Get the next iteration result steps go here, e.g.:
        const { value, done } = await ReadableStreamDefaultReaderRead(reader);
        if (done) {
          returnWasCalled = false;
          break;
        }
        yield value;
      }
    } catch (e) {
     // Async iterator return steps go here, with arg = e?
     } finally {
      // Async iterator return steps go here again, with arg = undefined?
    }
  }
}
```

I'm not so sure about that though, especially the return handling.

TimothyGu:
> (Nor is it especially good for sync iterators...)

In my defense, that's how the spec is [written](https://heycam.github.io/webidl/#dfn-value-pairs-to-iterate-over), and handles edge cases (e.g., values are deleted during iteration) just as well as the spec. I'm sure we can find a more efficient pattern if we tried hard enough, but that'd lead us to deviate from specs, so it's a tradeoff.

To give a sense of the complexity here, consider this transcript:
```js
const h = new Headers();
[...]

PR #224, opened 2020-05-18
Implement async iterables

domenic:
This starts to implement async iterables, #215, by following the spec as much as posisble.

This is missing:

- <del>Passing arguments to the `utils.asyncIteratorInit` hook. The hard part here is that I'm not sure how to implement [convert arguments for an async iterator method](https://heycam.github.io/webidl/#converting-arguments-for-an-asynchronous-iterator-method), since all our code to implement conversions and handle argument processing seems to be inside `generateOverloadConversions` which is pretty complicated.</del><ins>This wasn't so bad.</ins>

- Adding the `return` method appropriately. This is tricky because the Web IDL spec says to only add the return method "If an asynchronous iterator return algorithm is defined for the interface". That is, it's unclear how to generate this code without asking the runtime question of `utils.asyncIteratorReturn in Impl.prototype`. The best idea I have here is to add a new extended attribute, `[WebIDL2JSHasAsyncIteratorReturnSteps]`.

I would appreciate help here, especially on the former.


domenic:
OK, this is ready for review. I've integration-tested it with https://github.com/whatwg/streams/pull/1035 so the generated code works. But the source code could still benefit from review, and maybe we could pretty up the generated code. And if anyone has better ideas than `[WebIDL2JSHasReturnSteps]`, please let me know.

I put this on top of #225, but I don't think that's a strict dependency. I also tried to separate this out into semi-logical commits to show my progress and separate the parts I'm sure of from the parts I wasn't, but rebaselining the snapshot file is a pain, so that kind of stopped; it's probably best to just review the whole thing.

README.md
```diff
@@ -395,6 +395,20 @@ The second is the interface's supported property indices/names. By default, the
...
+
+For synchronous pair iterable declarations, the implementation class needs to implement the `[Symbol.iterator]()` property, returning an iterable of `[key, value]` pairs. These can be impls; the generated code will convert them into wrappers as necessary.
+
+### Async iterables
+
+[Asynchronous iterable declarations](https://heycam.github.io/webidl/#idl-async-iterable) require the implementation class to implement the following symbol-named methods, corresponding to algorithms from the Web IDL specification. The `utils` object below refers to the default export from the generated utilities file `utils.js`.
+
+- `utils.asyncIteratorNext`: corresponds to the [get the next iteration result](https://heycam.github.io/webidl/#dfn-get-the-next-iteration-result) algorithm, and receives a single argument containing an instance of the generated async iterator. For pair asynchronous iterables, the return value must be a `[key, value]` pair array, or `undefined` to signal the end of the iteration. For value asynchronous iterables, the return value must be the value, or `undefined` to signal the end of the iteration.
```
domenic, line 408: Hmm, this seems kind of bad. Streams can contain undefined values. Probably a spec bug?

TimothyGu: Web IDL doesn't really have a native “undefined” type though, so when the spec says "It must return a `Promise` that either rejects, resolves with undefined to signal the end of the iteration" in the IDL section (rather than the ES bindings section) I guess it's a bit unclear what exactly this "undefined" refers to. The conversion algorithms for [`any`](https://heycam.github.io/webidl/#es-any) says "an `object` reference to a special object that represents the ECMAScript **undefined** value".

domenic: I'll push a new commit that follows https://github.com/heycam/webidl/pull/885

domenic:
OK, this is ready for review. It continues to test well with the corresponding whatwg/streams PR.

lib/constructs/interface.js
```diff
@@ -416,7 +421,83 @@ class Interface {
...
+        this.str += `
+          ,
+          return(value) {
+            const internal = this[utils.iterInternalSymbol];
+            if (!internal) {
+              return Promise.reject(new TypeError("return() called on an object that is not an async iterator prototype object"));
+            }
+
+            if (internal.ongoingPromise) {
+              return Promise.reject(new TypeError("return() cannot be called while an ongoing call to next() has not settled"));
```
TimothyGu, line 484: Question: `ongoingPromise` is only set to `null` if the promise `next()` returns gets fulfilled, _not_ when it gets rejected. It seems to me that we might want the `[asyncIteratorReturn]` method to still be called (which probably cleans up some system resources) even if the last `next()` failed…?

lib/constructs/interface.js
```diff
@@ -416,7 +421,83 @@ class Interface {
...
+      this.str += `
+              });
+            };
+
+            internal.ongoingPromise = internal.ongoingPromise ? internal.ongoingPromise.then(nextSteps) : nextSteps();
+            return internal.ongoingPromise;
+          }
+      `;
+
+      if (this.iterable.hasReturnSteps) {
```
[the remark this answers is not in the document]
domenic: What do you mean by "have return added in the generated file"?

lib/constructs/interface.js
```diff
@@ -416,7 +421,83 @@ class Interface {
...
+        const AsyncIteratorPrototype = Object.create(utils.AsyncIteratorPrototype, {
+          [Symbol.toStringTag]: {
+            value: "${this.name} AsyncIterator",
+            configurable: true
+          }
+        });
+
+        Object.assign(AsyncIteratorPrototype, {
+          next() {
+            const internal = this[utils.iterInternalSymbol];
```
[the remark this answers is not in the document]
domenic: Note that sync iterables also have this issue; I went with their behavior. But I can fix both.

lib/constructs/interface.js
domenic: Good catch. I added a test to the streams WPT to catch this, and will send a Web IDL PR to fix it.

lib/constructs/interface.js
```diff
@@ -467,12 +467,15 @@ class Interface {
                 },
                 reason => {
                   internal.ongoingPromise = null;
+                  internal.isFinished = true;
```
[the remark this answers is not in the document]
domenic: I tested and we get

```js
[
  { status: 'fulfilled', value: { value: '0', done: false } },
  { status: 'fulfilled', value: { value: '1', done: false } },
  {
    status: 'rejected',
    reason: Error
        at TestImpl.[async iterator get the next iteration result] (C:\Users\Domenic\Dropbox\GitHub\jsdom\webidl2js\tmptest\Test-impl.js:12:29)
        at nextSteps (C:\Users\Domenic\Dropbox\GitHub\jsdom\webidl2js\tmptest\Test.js:30:79)
  },
  { status: 'fulfilled', value: { value: undefined, done: true } }
]
```

which seems as-desired.

lib/constructs/interface.js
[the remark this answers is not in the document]
domenic: I don't think that works because we only require the `Impl` class at the bottom of the file, I think for circular dependency reasons. At which time the `implementation` export might not exist.

PR #225, opened 2020-05-18
Tweak test cases for dictionary defaults

domenic:
In Web IDL these days, optional dictionary arguments must have a trailing "= {}". This adds those to the test .webidl files. Doing so uncovered an issue where DOMRectInit was referenced in the test files, but did not exist, so we add code to throw in this case, and correct the test file.

### DIFF
--- a/lib/parameters.js
+++ b/lib/parameters.js
@@ -18,7 +18,8 @@ function generateVarConversion(ctx, overload, i, parent, errPrefix, targetIdx = 
                                 overload.operation.arguments[i].default.type === "dictionary";
   if (isDefaultedDictionary && !ctx.dictionaries.has(idlType.idlType)) {
     throw new Error(
-      `The dictionary ${idlType.idlType} was referenced in an argument list, but doesn't exist`);
+      `The parameter ${overload.operation.arguments[i].name} was defaulted to {}, but no dictionary named ` +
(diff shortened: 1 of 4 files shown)


lib/parameters.js
```diff
@@ -14,7 +14,13 @@ function generateVarConversion(ctx, overload, i, parent, errPrefix, targetIdx =
   const requires = new utils.RequiresMap(ctx);
   const idlType = overload.typeList[i];
   // Always (try to) force-convert dictionaries
-  const optional = overload.optionalityList[i] === "optional" && !ctx.dictionaries.has(idlType.idlType);
+  const isDefaultedDictionary = overload.operation.arguments[i].default &&
+                                overload.operation.arguments[i].default.type === "dictionary";
+  if (isDefaultedDictionary && !ctx.dictionaries.has(idlType.idlType)) {
+    throw new Error(
+      `The dictionary ${idlType.idlType} was referenced in an argument list, but doesn't exist`);
```
TimothyGu, line 21: Couldn't it also be another class of errors: the default value is incorrectly set to be `{}`?

domenic: Fair point. So, maybe something like "The argument X was defaulted to {}, but there is no dictionary named Y".

PR #231, opened 2020-11-26
Avoid generating broken code when the `object` type is in a union

ExE-Boss:
**WebIDL2JS** currently crashes when [the `object` type][idl-object] is in a union, this fixes that.

[idl-object]: https://heycam.github.io/webidl/#idl-object

### DIFF
--- a/test/cases/NoUselessIfElse.webidl
+++ b/test/cases/NoUselessIfElse.webidl
@@ -1,0 +1,46 @@
+[Exposed=Window]
+interface NoUselessIfElse {
+  // # Overloads:
+  undefined overloadsObjectOrBoolean(object arg1);
+  undefined overloadsObjectOrBoolean(boolean arg1);
+
+  undefined overloadsObjectOrBooleanOrNumeric(object arg1);
+  undefined overloadsObjectOrBooleanOrNumeric(boolean arg1);
+  undefined overloadsObjectOrBooleanOrNumeric(long arg1);
+
+  undefined overloadsObjectOrNumeric(object arg1);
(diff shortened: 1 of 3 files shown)


test/cases/NoUselessIfElse.webidl
```diff
@@ -0,0 +1,46 @@
...
+  undefined overloadsObjectOrBooleanOrNumeric(boolean arg1);
+  undefined overloadsObjectOrBooleanOrNumeric(long arg1);
+
+  undefined overloadsObjectOrNumeric(object arg1);
+  undefined overloadsObjectOrNumeric(long arg1);
+
+  undefined overloadsBooleanOrNumeric(boolean arg1);
+  undefined overloadsBooleanOrNumeric(long arg1);
+
+  // ## Avoid regressions:
```
TimothyGu, line 17: I thought the present bug has to do with union types? Is overloading affected by the bug as well?

I'd also prefer a more surgical test suite. The snapshot generated from this file is really huge and not reviewable.

ExE-Boss: > Is overloading affected by the bug as well?

The current version isn’t, because overloads had the bug fixed at some point.

MattiasBuelens:
I solved this in a different way in #276, see https://github.com/jsdom/webidl2js/pull/276/commits/777c9aa9130fb24467db3cbf1836430b7917d1a2.

domenic:
Fixed in a different way as part of https://github.com/jsdom/webidl2js/pull/276.

PR #232, opened 2020-11-26
Avoid useless `if`‑`else` check during fallback type conversion

ExE-Boss:
This makes it so that for fallback types (e.g. `boolean` and `number`), if it’s the last fallback type among overloads or in a union, then the conversion will no longer be duplicated in the `if`‑and‑`else` branches:
```diff js
-} else if (typeof curArg === \\"boolean\\") {
-	{
-		let curArg = arguments[0];
-		curArg = conversions[\\"boolean\\"](curArg, {
-			context: \\"Failed to execute 'overloadsObjectOrBoolean' on 'NoUselessIfElse': parameter 1\\"
-		});
-		args.push(curArg);
-	}
 } else {
 	{
 		let curArg = arguments[0];
 		curArg = conversions[\\"boolean\\"](curArg, {
 			context: \\"Failed to execute 'overloadsObjectOrBoolean' on 'NoUselessIfElse': parameter 1\\"
 		});
 		args.push(curArg);
 	}
 }
```

## Depends on:
- [ ] <https://github.com/jsdom/webidl2js/pull/231>

### DIFF
--- a/lib/parameters.js
+++ b/lib/parameters.js
@@ -273,32 +273,41 @@ module.exports.generateOverloadConversions = function (ctx, typeOfOp, name, pare
         }
 
         const booleans = S.filter(o => isOrIncludes(ctx, o.typeList[d], t => t.idlType === "boolean"));
-        if (booleans.length) {
-          possibilities.push(`
-            if (typeof curArg === "boolean") {
-              ${continued(booleans[0], i)}
-            }
-          `);
-        }
-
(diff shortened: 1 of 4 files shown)


TimothyGu:
I'd rather not do this, since it makes the code harder to maintain by diverging from the [overload resolution algorithm](https://heycam.github.io/webidl/#es-overloads) in the spec.

PR #233, opened 2021-01-03
test: Add `NoUselessIfElse.webidl`

ExE-Boss:
Extracted from <https://github.com/jsdom/webidl2js/pull/231>


domenic:
This doesn't seem to address @TimothyGu's comment in #231 of

> I'd also prefer a more surgical test suite. The snapshot generated from this file is really huge and not reviewable.

domenic:
Fixed in a different way as part of https://github.com/jsdom/webidl2js/pull/276. (I apologize for holding this to such a high standard back in 2021; these days I'm more lenient.)

PR #234, opened 2021-03-01
Create sync iterators in the correct realm

ExE-Boss:
See <https://github.com/jsdom/jsdom/issues/2727> and <https://github.com/jsdom/jsdom/issues/2727#issuecomment-787559889>.

### DIFF
--- a/lib/constructs/interface.js
+++ b/lib/constructs/interface.js
@@ -412,25 +412,24 @@ class Interface {
     this.included.push(source);
   }
 
-  generateIterator() {
-    if (!this.iterable) {
+  generateInstallIteratorPrototype() {
+    const { iterable } = this;
+    if (!iterable) {
       return;
     }
 
-    if (this.iterable.isAsync) {
+    if (iterable.isAsync) {
       this.str += `
-        const AsyncIteratorPrototype = Object.create(utils.AsyncIteratorPrototype, {
+        ctorRegistry["${this.name} AsyncIterator"] = Object.assign(Object.create(utils.AsyncIteratorPrototype, {
           [Symbol.toStringTag]: {
             value: "${this.name} AsyncIterator",
             configurable: true
           }
-        });
-
-        Object.assign(AsyncIteratorPrototype, {
+        }, {
           next() {
             const internal = this && this[utils.iterInternalSymbol];
             if (!internal) {
-              return Promise.reject(new TypeError("next() called on a value that is not an async iterator prototype object"));
+              return Promise.reject(new TypeError("next() called on a value that is not a ${this.name} async iterator object"));
             }
 
             const nextSteps = () => {
@@ -446,7 +445,7 @@ class Interface {
                     internal.isFinished = true;
                     return { value: undefined, done: true };
                   }`;
-      if (this.iterable.isPair) {
+      if (iterable.isPair) {
         this.str += `
                   return utils.iteratorResult(next.map(utils.tryWrapperForImpl), kind);
         `;
@@ -472,12 +471,12 @@ class Interface {
           },
       `;
 
-      if (this.iterable.hasReturnSteps) {
+      if (iterable.hasReturnSteps) {
         this.str += `
           return(value) {
             const internal = this && this[utils.iterInternalSymbol];
             if (!internal) {
-              return Promise.reject(new TypeError("return() called on a value that is not an async iterator prototype object"));
+              return Promise.reject(new TypeError("return() called on a value that is not a ${this.name} async iterator object"));
             }
 
             const returnSteps = () => {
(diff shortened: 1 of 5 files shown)


lib/constructs/interface.js
```diff
@@ -420,25 +420,24 @@ class Interface {
...
           }
-        });
-
-        Object.assign(AsyncIteratorPrototype, {
+        }, {
           next() {
             const internal = this && this[utils.iterInternalSymbol];
             if (!internal) {
-              return Promise.reject(new TypeError("next() called on a value that is not an async iterator prototype object"));
+              return Promise.reject(new TypeError("next() called on a value that is not a WebIDL default async iterator object"));
```
domenic, line 440: These changes reduce clarity; web developers shouldn't need to know what Web IDL is.

lib/constructs/interface.js
```diff
@@ -1565,17 +1583,19 @@ class Interface {
...
           return;
         }
+
+        const ctorRegistry = utils.initCtorRegistry(globalObject);
     `;
 
     if (idl.inheritance) {
       this.str += `
-        if (globalObject.${idl.inheritance} === undefined) {
+        if (ctorRegistry.${idl.inheritance} === undefined) {
```
domenic, line 1592: Are these changes related to this PR?

lib/output/utils.js
```diff
@@ -12,6 +12,50 @@ const implSymbol = Symbol("impl");
...
+function isValidGlobalObject(globalObject) {
+  return intrinsicConstructors.every(intrinsic => typeof globalObject[intrinsic] === "function");
+}
+
+function initCtorRegistry(globalObject) {
+  if (hasOwn(globalObject, ctorRegistrySymbol)) {
+    return globalObject[ctorRegistrySymbol];
+  }
+
+  if (!isValidGlobalObject(globalObject)) {
```
domenic, line 34: I don't think this kind of checking is necessary.

lib/output/utils.js
```diff
@@ -12,6 +12,50 @@ const implSymbol = Symbol("impl");
...
+    throw new Error("Internal error: invalid global object");
+  }
+
+  const ctorRegistry = Object.create(
```
domenic, line 38: This is baroque. Instead use the simpler

```js
const ctorRegistry = Object.create(null);
for (const intrinsic of intrinsicConstructors) {
  ctorRegistryresult[`%${intrinsic}%`] = globalObject[intrinsic];
}
```

lib/output/utils.js
```diff
@@ -12,6 +12,50 @@ const implSymbol = Symbol("impl");
...
+      result[`%${intrinsic}%`] = {
+        enumerable: true,
+        value: globalObject[intrinsic]
+      };
+      return result;
+    }, Object.create(null))
+  );
+
+  // TODO: Also capture `%AsyncIteratorPrototype%`
+  Object.defineProperty(ctorRegistry, "%IteratorPrototype%", {
```
domenic, line 50: Again just use assignment.

lib/output/utils.js
```diff
@@ -12,6 +12,50 @@ const implSymbol = Symbol("impl");
...
+    enumerable: true,
+    value: Object.getPrototypeOf(Object.getPrototypeOf(new ctorRegistry["%Array%"]()[Symbol.iterator]()))
+  });
+
+  Object.defineProperty(globalObject, ctorRegistrySymbol, { value: ctorRegistry });
```
domenic, line 55: Just use assignment.

lib/constructs/interface.js
```diff
@@ -627,8 +635,18 @@ class Interface {
...
-            const iterator = Object.create(IteratorPrototype);
+          exports.createDefaultIterator = (globalObject, target, kind) => {
+            const ctorRegistry = globalObject[ctorRegistrySymbol];
+            if (ctorRegistry === undefined) {
+              throw new Error('Internal error: invalid global object');
+            }
+
+            if (ctorRegistry["${this.name}"] === undefined) {
+              throw new Error('Internal error: constructor ${this.name} is not installed on the passed global object');
+            }
```
domenic, line 646: I guess we should add these to stay consistent, but I am regretting them now; we should just remove all these checks and let runtime errors get thrown by attempting to access properties of undefined. They are not worth the generated code size.

Let's make sure we get them removed before the next release.

lib/output/utils.js
```diff
@@ -12,6 +12,50 @@ const implSymbol = Symbol("impl");
 const sameObjectCaches = Symbol("SameObject caches");
 const ctorRegistrySymbol = Symbol.for("[webidl2js]  constructor registry");
 
+const intrinsicConstructors = [
```
domenic, line 15: I think we only need Array for now? Can we remove the rest (while maybe adding a comment explaining that we're only adding what we need)?

domenic (changes requested):
Thanks for working on this; it seems like a good approach. There's some unrelated changes that don't belong in this PR, and some over-defensive coding that we should remove.

domenic (approved):

PR #238, opened 2021-03-07
Use Domenic's ESLint config as a base

domenic:


TimothyGu (approved):
Looks good. Do we want to start using Prettier or some other code formatter?

lib/types.js
```diff
@@ -83,7 +94,14 @@ function deepClone(obj) {
...
 }
 
-function generateTypeConversion(ctx, name, idlType, argAttrs = [], parentName, errPrefix = '"The provided value"') {
+function generateTypeConversion(
+  ctx,
+  name,
+  idlType,
+  argAttrs = [],
+  parentName = undefined,
+  errPrefix = '"The provided value"'
```
TimothyGu, line 103: I'm personally a fan of trailing commas too

domenic:
I'd be willing to use Prettier if it didn't diverge too much from my style (or could be configured as such). I haven't investigated...

lib/types.js
domenic: I can't bring myself to do it yet...

PR #239, opened 2021-03-14
Implement “internally create a new object implementing the interface”

ExE-Boss:
Extracted from <https://github.com/jsdom/webidl2js/pull/213> because of <https://github.com/jsdom/webidl2js/pull/213#issuecomment-792345633>:
> The generated snapshots here seem overcomplicated, and it's not clear why you're modifying existing functionality (like `new()`).

## Blocks:
- <https://github.com/jsdom/webidl2js/pull/213>

### DIFF
--- a/lib/constructs/interface.js
+++ b/lib/constructs/interface.js
@@ -1206,9 +1206,17 @@ class Interface {
 
   generateIface() {
     this.str += `
-      function makeWrapper(globalObject) {
-        const ctor = globalObject[ctorRegistrySymbol]["${this.name}"];
-        return Object.create(ctor.prototype);
+      function makeWrapper(globalObject, newTarget) {
+        let proto;
+        if (newTarget !== undefined) {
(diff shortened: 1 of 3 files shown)


lib/constructs/interface.js
```diff
@@ -1194,17 +1194,25 @@ class Interface {
 
   generateIface() {
     this.str += `
-      function makeWrapper(globalObject) {
+      function makeWrapper(globalObject, newTarget) {
```
domenic, line 1197: Sorry for the extra work, but please rebase on https://github.com/jsdom/webidl2js/pull/240.

domenic (commented):
Looks good after rebase

domenic (approved):
I'm sad about the extra size for such an edge-case feature, but it seems unavoidable, and I appreciate the work you did to make it shorter by using different steps than the spec (with equivalent logic).

PR #240, opened 2021-03-20
Remove extra checks from generated code

domenic:
These would often cause runtime failures anyway. Generating nicer error messages for such runtime failures is not worth the cost in generated code size, especially given that such scenarios indicate a coding failure on the part of the user of webidl2js, not the user of jsdom.


TimothyGu:
LGTM.

PR #242, opened 2021-03-23
Create errors in correct realm

ExE-Boss:
This makes it so that `TypeError`s and `Promise`s are created with the correct prototype chain.

### DIFF
--- a/lib/output/utils.js
+++ b/lib/output/utils.js
@@ -24,7 +24,14 @@ const sameObjectCaches = Symbol("SameObject caches");
 const ctorRegistrySymbol = Symbol.for("[webidl2js]  constructor registry");
 
 // This only contains the intrinsic names that are referenced from the `ctorRegistry`:
-const intrinsicConstructors = ["Array"];
+const intrinsicConstructors = [
+  "Array",
+  // Number and String are used by `webidl-conversions`
+  "Number",
+  "Promise",
+  "String",
+  "TypeError"
+];
 
 function initCtorRegistry(globalObject) {
   if (hasOwn(globalObject, ctorRegistrySymbol)) {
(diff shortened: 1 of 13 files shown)


test/__snapshots__/test.js.snap
```diff
@@ -6,9 +6,15 @@ exports[`generation built-in types Function 1`] = `
 const conversions = require(\\"webidl-conversions\\");
 const utils = require(\\"./utils.js\\");
 
-exports.convert = (value, { context = \\"The provided value\\" } = {}) => {
+exports.convert = (
+  value,
+  {
+    context = \\"The provided value\\",
+    ctorRegistry = { \\"%TypeError%\\": TypeError, \\"%Number%\\": Number, \\"%Promise%\\": Promise, \\"%String%\\": String }
```
domenic, line 13: Hmm. I think it'd be a bit more uniform for consumers if the convert algorithm took a mandatory first-parameter global, similar to `create()` and friends.

lib/output/utils.js
```diff
@@ -45,6 +52,16 @@ function initCtorRegistry(globalObject) {
   return ctorRegistry;
 }
 
+// The result is used as the `globals` property of the `options` parameter for `webidl-conversions`:
+function conversionGlobalsFromCtorRegistry(ctorRegistry) {
```
domenic, line 56: Let's stop storing the intrinsics under `"%TypeError"` and instead store them under `"TypeError"`. Then we can get rid of this function entirely and just pass the `ctorRegistry`.

domenic (commented):
Although this isn't technically correct (as it uses the relevant realm, not the current realm), this is an exciting step in the right direction. Although we might, at some point, have to undo this work if we do something like we talked about in https://github.com/jsdom/jsdom/issues/2727, it's still worthwhile in the meantime, I think.

PR #251, opened 2021-09-12
Create exceptions and promises in the correct realm

domenic:
Closes #242 by superseding it.

This approach moves the ctorRegistry back to being mostly for the registration of WebIDL2JS-generated constructors. We use the globalObject variable to pass to webidl-conversions and to find our Promise and TypeError constructors.

Needs testing with jsdom before merging.


domenic:
Review appreciated if either of you have time, @TimothyGu @ExE-Boss. https://github.com/jsdom/webidl2js/commit/8faa068ac83755f0294994e0d256108b790cccf3 in particular I am unsure what is going on.

lib/output/utils.js
```diff
@@ -29,10 +29,7 @@ function newObjectInRealm(globalObject, object) {
 const wrapperSymbol = Symbol("wrapper");
 const implSymbol = Symbol("impl");
 const sameObjectCaches = Symbol("SameObject caches");
-const ctorRegistrySymbol = Symbol.for("[webidl2js]  constructor registry");
-
-// This only contains the intrinsic names that are referenced from the `ctorRegistry`:
-const intrinsicConstructors = ["Array"];
```
ExE-Boss, line 35: Storing these in the constructor registry was done to correctly implement the parts of **WebIDL** that use intrinsic objects, and to allow using `extends ctorRegistry.<base‑name>` instead of `extends globalObject.<base‑name>`.

lib/output/utils.js
```diff
@@ -42,13 +39,13 @@ function initCtorRegistry(globalObject) {
   }
 
   const ctorRegistry = Object.create(null);
-  for (const intrinsic of intrinsicConstructors) {
-    ctorRegistry[`%${intrinsic}%`] = globalObject[intrinsic];
```
ExE-Boss, line 46: Refs <https://github.com/jsdom/webidl2js/pull/242#discussion_r606838290>:
```suggestion
    ctorRegistry[intrinsic] = globalObject[intrinsic];
```

ExE-Boss (changes requested):

lib/output/utils.js
```diff
@@ -29,10 +29,7 @@ function newObjectInRealm(globalObject, object) {
 const wrapperSymbol = Symbol("wrapper");
 const implSymbol = Symbol("impl");
 const sameObjectCaches = Symbol("SameObject caches");
-const ctorRegistrySymbol = Symbol.for("[webidl2js]  constructor registry");
-
-// This only contains the intrinsic names that are referenced from the `ctorRegistry`:
-const intrinsicConstructors = ["Array"];
+const ctorRegistrySymbol = Symbol("[webidl2js] constructor registry");
```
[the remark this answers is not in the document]
domenic: Hmm, I see. Probably then the problem is that I was testing only upgrading webidl2js for jsdom, but not domexception and whatwg-url. I will have to test upgrading all three. Or maybe it would be best to add a major version number to the symbol description so that if we do mix versions it doesn't cause problems.

lib/output/utils.js
domenic: I think we still implement the parts Web IDL that use intrinsic objects correctly. And I don't see the value of `extends ctorRegistry.` instead of `extend globalObject.`; indeed requiring everything to use the same ctorRegistry seems subpar instead of using the global as the coordination point.

lib/output/utils.js
domenic: You are commenting on a deleted line.

PR #255, opened 2022-01-23
Avoid "TypeError: Cannot convert undefined or null to object" in Angular Universal

vkarpov15:
Sorry for the somewhat odd PR, but this issue is making the MongoDB Node driver and Mongoose fail in Angular Universal. Angular Universal does some strange transpiling that breaks how this lib is getting AsyncIteratorPrototype.

I tracked this down from https://github.com/Automattic/mongoose/issues/11155, https://github.com/angular/angular-cli/issues/21735, and opened a PR to work around this issue in Angular here: https://github.com/angular/angular-cli/pull/22550.

### DIFF
--- a/lib/output/utils.js
+++ b/lib/output/utils.js
@@ -31,7 +31,8 @@ const implSymbol = Symbol("impl");
 const sameObjectCaches = Symbol("SameObject caches");
 const ctorRegistrySymbol = Symbol.for("[webidl2js] constructor registry");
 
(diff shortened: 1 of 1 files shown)


domenic:
Sounds like a bug in their transpiler; this is not the right place to change.

vkarpov15:
Angular Universal team says that their transpiler is "correct" :( https://github.com/angular/angular-cli/issues/21735#issuecomment-915974013 Whether it actually is, is open to debate.

I think this PR is sufficiently low risk and unobtrusive to merge and unblock users. Can you please reconsider?

Issue #113, opened 2018-08-28
Throws error when shadowing inherited getter


Zirro:
The following combination of IDLs currently result in a `duplicated named getter`-error:

```webidl
[Exposed=Window, LegacyUnenumerableNamedProperties]
interface HTMLCollection {
  readonly attribute unsigned long length;
  [WebIDL2JSValueAsUnsupported=null] getter Element? item(unsigned long index);
  [WebIDL2JSValueAsUnsupported=null] getter Element? namedItem(DOMString name);
};

[Exposed=Window]
interface HTMLFormControlsCollection : HTMLCollection {
  // inherits length and item()
  [WebIDL2JSValueAsUnsupported=null] getter (RadioNodeList or Element)? namedItem(DOMString name); // shadows inherited namedItem()
};
```

`Error: Invalid getter "namedItem" on interface HTMLFormControlsCollection (defined in HTMLCollection): duplicated named getter`

The `namedItem()` on `HTMLFormControlsCollection` shadows the one from `HTMLCollection`. Going by the spec, it seems like this should be possible.

TimothyGu:
The Web IDL spec [says](https://heycam.github.io/webidl/#idl-special-operations):

> If an object implements more than one [interface](https://heycam.github.io/webidl/#dfn-interface) that defines a given special operation, then it is undefined which (if any) special operation is invoked for that operation.

That was my original reason for disallowing shadowing.

domenic:
Moved to https://github.com/heycam/webidl/issues/833

UndefinedBehaviour:
I understand the root of the issue is in the webidl spec, but I think it would make sense to implement what browsers do (I checked Chrome and Firefox) i.e. use the definition in the most derived class.
This would allow to implement HTMLFormControlsCollection (and do away with one [TODO](https://github.com/jsdom/jsdom/blob/master/lib/jsdom/living/nodes/HTMLFormElement-impl.js#L64) in the code), which in turn would allow to implement RadioNodeList (and close https://github.com/jsdom/jsdom/issues/2600).
What do you think?

domenic:
Yeah, a PR to fix this with a link to the relevant spec issue would be welcome.

PR #258, opened 2022-02-04
allow getters, setters and deleters to be shadowed in derived class

UndefinedBehaviour:
https://github.com/jsdom/webidl2js/issues/113

### DIFF
--- a/lib/constructs/interface.js
+++ b/lib/constructs/interface.js
@@ -213,9 +213,6 @@ class Interface {
             }
             this.indexedGetter = member;
           } else if (isNamed(member)) {
-            if (this.namedGetter) {
-              throw new Error(`${msg}duplicated named getter`);
-            }
             this.namedGetter = member;
           } else {
             throw new Error(`${msg}getter is neither indexed nor named`);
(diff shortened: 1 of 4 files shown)


domenic:
Thanks! Can you add a test by dropping a file or two into https://github.com/jsdom/webidl2js/tree/master/test/cases , which fails before this change but passes after?

UndefinedBehaviour:
There you go!

lib/constructs/interface.js
```diff
@@ -213,9 +213,6 @@ class Interface {
             }
             this.indexedGetter = member;
           } else if (isNamed(member)) {
-            if (this.namedGetter) {
-              throw new Error(`${msg}duplicated named getter`);
-            }
```
UndefinedBehaviour, line 218: You could also change the condition to `if (this.hasOwnProperty(member))` in order to keep at least a check on duplicated member names, but I noticed this kind of check is not in place for regular attributes either, so i preferred to have the same behaviour for getters and setters as well i.e. if multiple members have the same name, simply use the last defined one and discard all the others.

domenic (approved):

PR #272, opened 2025-04-20
Fix generated set trap for proxy cases

domenic:
Fixes https://github.com/jsdom/jsdom/issues/3565.

### DIFF
--- a/lib/constructs/interface.js
+++ b/lib/constructs/interface.js
@@ -582,7 +582,7 @@ class Interface {
   generateExport() {
     this.str += `
       exports.is = value => {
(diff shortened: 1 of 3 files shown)


dverbru:
Is this going to be included in the next jsdom v26 patch release? Thank you

domenic:
It is automatically; jsdom uses `^` in its version specifiers so any installations automatically get patch version updates.

domenic:
You're right, it's in the build step. But the good news is that jsdom 26.1.0 was built with the version 18.0.0, so the regression in 18.0.1 which this commit fixes never made it into jsdom.

PR #276, opened 2025-09-25
Support `[AllowResizable]`, `[AllowShared]`, `SharedArrayBuffer`, and `AllowSharedBufferSource`

MattiasBuelens:
* `SharedArrayBuffer` is now a separate IDL type, and `[AllowShared] BufferSource` has been replaced with a new `AllowSharedBufferSource` typedef.
  * See: https://github.com/whatwg/webidl/pull/1311
* Buffer source types can now have extended attributes such as `[AllowResizable]` and `[AllowShared]`, which need to be checked by [`webidl-conversions`](https://github.com/jsdom/webidl-conversions/). Thus, we now generate the appropriate conversion calls.
  * See: https://github.com/whatwg/webidl/pull/982 and https://github.com/jsdom/webidl-conversions/pull/51.
* Fix unions with callback functions and callback interfaces.
* Simplify generated code for `object` type in unions.
  * This also fixes the crash reported in #231.

### DIFF
--- a/lib/types.js
+++ b/lib/types.js
@@ -202,9 +202,13 @@ function generateTypeConversion(
     }
 
     if (union.object) {
-      output.push(`if (utils.isObject(${name}) && ${name}[utils.implSymbol]) {
-                     ${name} = utils.implForWrapper(${name});
-                   }`);
+      output.push(`
+        if (utils.isObject(${name})) {
+          if (${name}[utils.implSymbol]) {
+            ${name} = utils.implForWrapper(${name});
+          }
+        }
+      `);
     } else if (union.interfaces.size > 0) {
       const exprs = [...union.interfaces].map(iface => {
         let fn;
@@ -224,17 +228,17 @@ function generateTypeConversion(
       `);
     }
 
-    if (union.ArrayBuffer || union.object) {
+    if (union.ArrayBuffer) {
       output.push(`if (utils.isArrayBuffer(${name})) {
         ${generateTypeConversion(ctx, name, union.ArrayBuffer, [], parentName, errPrefix).body}
       }`);
     }
-    if (union.SharedArrayBuffer || union.object) {
+    if (union.SharedArrayBuffer) {
       output.push(`if (utils.isSharedArrayBuffer(${name})) {
         ${generateTypeConversion(ctx, name, union.SharedArrayBuffer, [], parentName, errPrefix).body}
       }`);
     }
-    if (union.ArrayBufferViews.size > 0 || union.object) {
+    if (union.ArrayBufferViews.size > 0) {
       const viewIdlTypes = new Set([...union.ArrayBufferViews].map(item => item.idlType));
       // Skip specific type check if all ArrayBufferView member types are allowed.
       if (viewIdlTypes.size === arrayBufferViewTypes.size) {
@@ -253,30 +257,24 @@ function generateTypeConversion(
       }
     }
 
-    if (union.callbackFunction || union.object) {
-      let code = `if (typeof ${name} === "function") {`;
-
-      if (union.callbackFunction) {
-        const conv = generateTypeConversion(
-          ctx,
-          name,
-          union.callbackFunction,
-          [],
-          parentName,
-          `${errPrefix} + " callback function"`
-        );
-        requires.merge(conv.requires);
-        code += conv.body;
-      } else if (union.object) {
-        // noop
-      }
-
-      code += "}";
-
-      output.push(code);
+    if (union.callbackFunction) {
+      const conv = generateTypeConversion(
+        ctx,
+        name,
+        union.callbackFunction,
+        [],
+        parentName,
+        `${errPrefix} + " callback function"`
+      );
+      requires.merge(conv.requires);
+      output.push(`
+        if (typeof ${name} === "function") {
+          ${conv.body}
+        }
+      `);
     }
 
-    if (union.sequenceLike || union.dictionary || union.record || union.object || union.callbackInterface) {
+    if (union.sequenceLike || union.dictionary || union.record || union.callbackInterface) {
       let code = `if (utils.isObject(${name})) {`;
 
       if (union.sequenceLike) {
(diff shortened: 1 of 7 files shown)


lib/types.js
```diff
@@ -214,18 +224,43 @@ function generateTypeConversion(
...
-      if (union.ArrayBufferViews.size !== arrayBufferViewTypes.size) {
-        const exprs = [...union.ArrayBufferViews].map(a => `${name}.constructor.name === "${a}"`);
-        condition += ` && (${exprs.join(" || ")})`;
+      if (union.ArrayBufferViews.size > 0) {
+        const viewIdlTypes = new Set([...union.ArrayBufferViews].map(item => item.idlType));
+        // Skip specific type check if all ArrayBufferView member types are allowed.
+        if (viewIdlTypes.size === arrayBufferViewTypes.size) {
+          const viewExtAttrs = mergeExtAttrsOfTypes(union.ArrayBufferViews);
+          output.push(`if (ArrayBuffer.isView(${name})) {
+            ${generateWebIDLConversions(`conversions["ArrayBufferView"]`, viewExtAttrs)}
```
MattiasBuelens: This bit is slightly awkward. I can't use `generateTypeConversion` with IDL type `ArrayBufferView`, since that will just resolve the union again and recurse back to this same line. So I have to jump straight into `generateWebIDLConversions` instead.

test/cases/TypedefsAndUnions.webidl
```diff
@@ -4,11 +4,13 @@ interface TypedefsAndUnions {
   undefined numOrEnumConsumer((double or RequestDestination)? a);
   undefined numOrStrOrNullConsumer(NumOrStrOrNull a);
   undefined numOrStrOrURLOrNullConsumer(NumOrStrOrURLOrNull? a);
+  undefined numOrObjConsumer((double or object) a);
```
MattiasBuelens, line 7: There wasn't a test case yet for a union that includes `object`, which turns out to be a little tricky! [The spec](https://webidl.spec.whatwg.org/#js-union) says:
> 6. If V [is an Object](https://tc39.es/ecma262/#sec-object-type), V has an [[ArrayBufferData]] [internal slot](https://tc39.es/ecma262/#sec-object-internal-methods-and-internal-slots), and [IsSharedArrayBuffer](https://tc39.es/ecma262/multipage/structured-data.html#sec-issharedarraybuffer)(V) is false, then:
>    1. If types includes [ArrayBuffer](https://webidl.spec.whatwg.org/#idl-ArrayBuffer), then return the result of [converting](https://webidl.spec.whatwg.org/#dfn-convert-ecmascript-to-idl-value) V to [ArrayBuffer](https://webidl.spec.whatwg.org/#idl-ArrayBuffer).
>    2. If types includes [object](https://webidl.spec.whatwg.org/#idl-object), then return the IDL value that is a reference to the object V.

We must not run any conversions on the `ArrayBuffer`, instead we must keep it as-is. That's why we generate this code:
```javascript
let curArg = arguments[0];
if (utils.isObject(curArg) && curArg[utils.implSymbol]) {
  curArg = utils.implForWrapper(curArg);
} else if (utils.isArrayBuffer(curArg)) {
} else if (utils.isSharedArrayBuffer(curArg)) {
} else if (ArrayBuffer.isView(curArg)) {
} else if (typeof curArg === "function") {
} else if (utils.isObject(curArg)) {}
} else { ... }
```

MattiasBuelens:
@domenic Could you assign some reviewer(s) to this PR? I can't seem to do it myself.

domenic:
I'll still work on jsdom stuff, don't worry. I plan to review this tomorrow :)

test/cases/TypedefsAndUnions.webidl
domenic: How hard would it be to make that nicer, i.e.

```js
let curArg = arguments[0];
if (utils.isObject(curArg)) {
  if (curArg[utils.implSymbol]) {
    curArg = utils.implForWrapper(curArg);
  }
} else { ... }
```

?

If it's too annoying to do right now, adding an explanatory comment for all the empty if blocks would be good. (In the generator code, not the generated code.)

lib/types.js
domenic: This doesn't seem that bad to me, but a comment would be helpful.

domenic:
I appreciate the separate commits; it made reviewing this easier. If you're up for some rebase-fu, here are some minor notes:

- "Rename operations" -> "Rename test operations for clarity"
- Merged "Fix object type handling" and "Test object inside a union", and expand on the fix (maybe with reference to #231).
- Merge "Add AllowSharedBufferSource" into "Support [AllowShared] on all ArrayBufferView types"
- Edit the commit message for "Generate conversions for ArrayBufferView types" to explain why we're doing this now when we didn't before.

Or, if you prefer, we can just do a big squash-and-merge with a sufficiently-detailed commit message.

test/cases/TypedefsAndUnions.webidl
MattiasBuelens: This is following [the specification for converting a union type](https://webidl.spec.whatwg.org/#js-union) pretty much to the letter: we emit a check for every case of "if *types* includes `object`" (i.e. if `union.object` is set).

But looking at `extractUnionInfo` and the rules for [distinguishable types](https://webidl.spec.whatwg.org/#dfn-distinguishable), you're not allowed to have `(ArrayBuffer or object)`, so that throws during code generation. So I think your suggestion is valid, and we can eliminate [these `|| union.object` branches in `generateUnion`](https://github.com/jsdom/webidl2js/blob/24bd5987397ef62cd5eb78132aadfe678fb90787/lib/types.js#L218).

domenic (approved):
Awesome stuff, thank you!

The repeated conversions for unions with primitives are a bit strange, but I see the follow the spec exactly. (I.e. https://webidl.spec.whatwg.org/#js-union steps 13.1 + 17.)

Probably some optimizations are possible for simple unions. But I think the right tactic there would be to work on the more general project of slimming down the generated code, driven by analyzing all of jsdom to see what are the worst offenders weighted by web platform prevalence.

PR #278, opened 2026-04-13
Support async sequence types

MattiasBuelens:
This PR adds support for [async sequence types](https://webidl.spec.whatwg.org/#idl-async-iterable-type). It was first introduced as `async iterable<T>` in https://github.com/whatwg/webidl/pull/1397, and later renamed to `async_sequence<T>` in https://github.com/whatwg/webidl/pull/1500. (Note that this is different from [_asynchronously iterable declarations_](https://webidl.spec.whatwg.org/#dfn-async-iterable-declaration), which are already supported since #224.)

The implementation is largely based on the Streams reference implementation (https://github.com/whatwg/streams/pull/1083). Once this PR lands, we will update Streams to use webidl2js's version instead.

### DIFF
--- a/lib/output/utils.js
+++ b/lib/output/utils.js
@@ -226,6 +226,9 @@ function createAsyncFromSyncIterator(syncIterator) {
 }
 
 function convertAsyncSequence(object, itemConverter, errPrefix = "The provided value") {
+  if (!isObject(object)) {
+    throw new TypeError(`${errPrefix} is not an object.`);
+  }
   let method = getMethod(object, Symbol.asyncIterator, errPrefix);
   let type = "async";
   if (method === undefined) {
--- a/lib/output/utils.js
+++ b/lib/output/utils.js
@@ -325,6 +325,7 @@ module.exports = exports = {
   isArrayBuffer,
   isSharedArrayBuffer,
   isArrayIndexPropName,
+  getMethod,
   convertAsyncSequence,
   supportsPropertyIndex,
   supportedPropertyIndices,
--- a/lib/output/utils.js
+++ b/lib/output/utils.js
@@ -204,7 +204,7 @@ function getMethod(value, property, errPrefix = "The provided value") {
     return undefined;
   }
   if (typeof func !== "function") {
-    throw new TypeError(`${errPrefix} is not a function.`);
+    throw new TypeError(`${errPrefix}'s ${property} property is not a function.`);
   }
   return func;
 }
@@ -226,10 +226,10 @@ function createAsyncFromSyncIterator(syncIterator) {
 }
 
 function convertAsyncSequence(object, itemConverter, errPrefix = "The provided value") {
-  let method = getMethod(object, Symbol.asyncIterator, `${errPrefix}'s [Symbol.asyncIterator] method`);
+  let method = getMethod(object, Symbol.asyncIterator, errPrefix);
   let type = "async";
   if (method === undefined) {
-    method = getMethod(object, Symbol.iterator, `${errPrefix}'s [Symbol.iterator] method`);
+    method = getMethod(object, Symbol.iterator, errPrefix);
     if (method === undefined) {
       throw new TypeError(`${errPrefix} is not an async iterable object.`);
     }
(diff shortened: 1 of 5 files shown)


MattiasBuelens:
@domenic Could you review, or assign someone to review? Thanks. 🙏

lib/types.js
```diff
@@ -278,14 +281,21 @@ function generateTypeConversion(
       let code = `if (utils.isObject(${name})) {`;
 
       if (union.sequenceLike) {
-        code += `if (${name}[Symbol.iterator] !== undefined) {`;
+        if (union.sequenceLike.generic === "async_sequence") {
+          code += `if (
+            ${name}[Symbol.asyncIterator] !== undefined ||
+            ${name}[Symbol.iterator] !== undefined
```
MattiasBuelens: This is *supposed* to use `GetMethod` which throws for non-functions, as per [step 11.1.1.1](https://webidl.spec.whatwg.org/#js-union). But for sequences and frozen arrays, we seem to just check for `undefined`? Should we fix this?

domenic: I think we should fix this, and ideally have some tests.

lib/types.js
```diff
@@ -775,8 +812,12 @@ function areDistinguishable(ctx, type1, type2) {
     const isDictionaryLike2 = ctx.dictionaries.has(inner2.idlType) ||
                               ctx.callbackInterfaces.has(inner2.idlType) ||
                               inner2.generic === "record";
-    const isSequenceLike1 = inner1.generic === "sequence" || inner1.generic === "FrozenArray";
-    const isSequenceLike2 = inner2.generic === "sequence" || inner2.generic === "FrozenArray";
+    const isSequenceLike1 = inner1.generic === "sequence" ||
```
domenic: Can you factor out an `isSequenceLike(idleType)` helper, so we can call this as `isSequenceLike(inner1)`?

lib/output/utils.js
```diff
@@ -198,6 +198,103 @@ function isAccessorDescriptor(desc) {
...
+  };
+}
+
+function openAsyncSequence(object, method, type, itemConverter, errPrefix = "The provided value") {
+  let iterator = call(method, object);
+  if (!isObject(iterator)) {
+    throw new TypeError(`${errPrefix}'s method must return an object`);
+  }
+  let nextMethod = iterator.next;
+  if (typeof nextMethod !== "function") {
```
domenic: I can't actually find this step in the spec. I think we're supposed to store the nextMethod and fail when calling it. (Again, tests would be nice. But definitely not mandatory...)

domenic (commented):
Can you add a test case (just a snapshot test)?

lib/output/utils.js
MattiasBuelens: Huh, indeed. [`GetIteratorDirect`](https://tc39.es/ecma262/multipage/abstract-operations.html#sec-getiteratordirect) only does `Get(obj, "next")`, not `GetMethod(obj, "next")`.

I'll remove this.

lib/types.js
MattiasBuelens: Agreed, this need more tests. This is tricky stuff! 😅

I think I'll also port some of the input validation tests from [`readable-streams/from.any.js`](https://github.com/web-platform-tests/wpt/blob/48fe2a8e29d44b7764aea192e84c8eb931f36ed6/streams/readable-streams/from.any.js) over to this project.

domenic: I'm OK with this project not having any runtime tests, but I'm just thinking about if this isn't covered by anything in WPT, then it's good to improve WPT coverage.

lib/types.js
```diff
@@ -362,6 +372,33 @@ function generateTypeConversion(
...
+      "item",
+      idlType.idlType[0],
+      [],
+      parentName,
+      `${errPrefix} + "'s element"`
+    );
+    requires.merge(conv.requires);
+
+    str += `
+      if (!utils.isObject(${name})) {
```
domenic: Would it make more sense to move the isObject() test inside convertAsyncSequence()? Then the generated code contains a bit less duplicated code.

domenic (approved):
LGTM with one question; after hearing back on that, I will merge and release unless you say otherwise.

lib/types.js
MattiasBuelens: I'll have a look around. [The `URLSearchParams` constructor](https://url.spec.whatwg.org/#interface-urlsearchparams) takes a union with a sequence type, maybe we can add a test there. Or perhaps the IDL harness tests already cover this case?

domenic (approved):

PR #285, opened 2026-09-05
Cache type checks for generated proxies

scttcper:
Repeated jsdom collection and document checks go through Proxy traps to read the implementation symbol. Register generated proxies in a WeakSet and return early for known wrappers; unfamiliar objects still use the existing validation.

A React list with 500 rows, 20 rerenders, and RTL text queries measured:

| CPU per workload | Main | This PR |
|---|---:|---:|
| First run | 315.9 ms | 295.9 ms |
| Warmed | 142.6 ms | 126.7 ms |

Seven alternating fresh-process pairs, comparing generated jsdom bindings from current webidl2js main against this change. All seven warmed pairs improved; median paired CPU reduction was 12.1% (95% bootstrap interval: 6.5–18.4%). First-run times exclude module loading.

The tradeoff is about 1 MiB of additional retained heap with 30,000 collection wrappers. Weak-reference probes confirmed wrappers and old documents can still be collected.

Related to #284, but this does not change validation of user-created proxies. Most of the diff is regenerated snapshots; the generator change is in `lib/constructs/interface.js`.

### DIFF
--- a/lib/constructs/interface.js
+++ b/lib/constructs/interface.js
@@ -581,7 +581,15 @@ class Interface {
 
   generateExport() {
     this.str += `
+      ${this.isLegacyPlatformObj ? "const knownWrappers = new WeakSet();" : ""}
       exports.is = value => {
+        ${this.isLegacyPlatformObj ?
+          `
+          if (knownWrappers.has(value)) {
+            return true;
+          }
+        ` :
+          ""}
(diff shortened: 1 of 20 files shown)


domenic:
I actually have my own on-disk branch comparing several approaches here that I was going to put up, which improves spec conformance while also improving perf... I'll add this to the bakeoff and have Codex figure out what the best path is.

